PR #22, opened 2022-03-30
feat: use HTTP status codes in superagent-codec-adapter

bitgopatmcl:
Result of discussion in #18 

Ticket: BG-45933


bitgopatmcl:
Re-opening because GH does not seem to be picking up my rebased commits in this PR

PR #25, opened 2022-03-30
feat: use HTTP status codes in superagent-codec-adapter

bitgopatmcl:
Ticket: BG-45933


ericcrosson-bitgo (approved):
Coupling the superagent adapter to the HTTP protocol and returning standard HTTP status codes

PR #26, opened 2022-03-30
ci: reduce runtime and upgrade GitHub actions

ericcrosson-bitgo:
- upgrade setup-node to v3
- cache npm dependencies between builds
- use npm ci over npm install


bitgopatmcl (approved):
Speeds up CI

PR #27, opened 2022-04-01
feat: simplify intersections of simple types

bitgopatmcl:
This is mostly noticeable with `optionalize`d codecs as they internally
are represented as an `io-ts` intersection of two types, one containing
the required properties, and the other the optional ones. With this
change, the resulting OpenAPI spec won't needlessly generate an `allOf`
for these.

Ticket: BG-46081


ericcrosson-bitgo (approved):
Looks great

PR #29, opened 2022-04-01
feat: add shell.nix

bitgopatmcl:
This one is very minimal, but it works

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,14 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+let
+  PROJECT_ROOT=builtins.toString ./.;
+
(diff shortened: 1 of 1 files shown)


bitgopatmcl:
Node 12 CI fixed in #30

ericcrosson-bitgo (approved):
What about an `.envrc` file?

bitgopatmcl:
> What about an .envrc file?

That would require us to decide to standardize on using `lorri` or possibly some similar plugin

ericcrosson-bitgo:
I could get behind that

EricCrosson:
Just seen https://github.com/nix-community/nix-direnv

PR #30, opened 2022-04-01
fix: target es2019 for Node 12 support

bitgopatmcl:


ericcrosson-bitgo (approved):
Is the reduction of `target` to `ES2019` required to support Node.js 12?

bitgopatmcl:
Looks like it: https://stackoverflow.com/questions/59787574/typescript-tsconfig-settings-for-node-js-12

PR #31, opened 2022-04-01
Package rename

bitgopatmcl:


packages/express-wrapper/tsconfig.json
```diff
@@ -29,7 +29,7 @@
       "path": "../io-ts-http"
     },
     {
-      "path": "../superagent-codec-adapter"
+      "path": "../superagent-wrapper"
```
ericcrosson-bitgo, line 32: I presume you made these edits manually? I was writing documentation in https://github.com/typescript-tools/spec this week and pondering the changes required to support npm 7 workspaces without lerna in the `link` command

ericcrosson-bitgo (approved):

ericcrosson-bitgo:
Rename of packages looks good to me

packages/express-wrapper/tsconfig.json
bitgopatmcl: Yeah, I did it by hand. I wonder if turborepo has a command to dump the dependency graph?

ericcrosson-bitgo: It does, into `dot`, at least

PR #32, opened 2022-04-05
Prepare for continuous deployments

ericcrosson-bitgo:
- remove unused dependencies
- remove placeholder npm script


bitgopatmcl (approved):
Removes unused dependencies, and removing the `test` script doesn't appear to make turborepo complain.

PR #34, opened 2022-04-05
fix: inference of `ResponseType<HttpRoute>`

bitgopatmcl:
Instead of `never` this should be something like the union of all
responses with `any` or `unknown` body.

Broken:
image

Fixed:
image

### DIFF
--- a/packages/io-ts-http/src/httpResponse.ts
+++ b/packages/io-ts-http/src/httpResponse.ts
@@ -2,11 +2,13 @@ import * as t from 'io-ts';
 
 import { Status } from '@api-ts/response';
 
-export type HttpResponse = t.Props;
+export type HttpResponse = {
+  [K: string]: t.Mixed;
+};
 
-export type KnownResponses<Response extends HttpResponse> = {
-  [K in keyof Response]: K extends Status ? K : never;
-}[keyof Response];
+type KnownResponses<Response extends HttpResponse> = {
+  [K in Status]: K extends keyof Response ? K : never;
+}[Status];
 
 export const HttpResponseCodes: { [K in Status]: number } = {
   ok: 200,
--- a/packages/io-ts-http/src/httpResponse.ts
+++ b/packages/io-ts-http/src/httpResponse.ts
@@ -3,14 +3,18 @@ import * as t from 'io-ts';
 import { Status } from '@api-ts/response';
 
 export type HttpResponse = {
-  [K: string]: t.Mixed;
+  [K in Status]?: t.Mixed;
 };
 
-type KnownResponses<Response extends HttpResponse> = {
-  [K in Status]: K extends keyof Response ? K : never;
-}[Status];
+export type KnownResponses<Response extends HttpResponse> = {
+  [K in keyof Response]: K extends Status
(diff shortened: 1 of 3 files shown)


bitgopatmcl:
This is primarily for correctness. The intention of `ResponseType<thing>` is to be used on a subtype of `HttpResponse`, but the boundary case should work.

ericcrosson-bitgo:
I scheduled a meeting to discuss this change in more detail; I think I'm too hazy on the specifics but I have a concern I'd love to run past you in a higher-bandwidth medium before approving

packages/superagent-wrapper/src/request.ts
```diff
@@ -142,19 +151,16 @@ const patchRequest = <Req extends SuperAgentRequest, Route extends h.HttpRoute>(
...
     patchedReq.decode().then((res) => {
       if (res.status !== status) {
         const error = res.error ?? `Unexpected status code ${res.status}`;
         throw new Error(JSON.stringify(error));
       } else {
-        return {
-          body: res.body,
-          original: res.original,
-        };
+        return res as ExpectedDecodedResponse<Route, StatusCode>;
```
bitgopatmcl, line 163: To be honest I actually don't know why the type of `res` isn't narrowed by control flow here. Maybe just because of some complexity limit?

packages/io-ts-http/src/httpResponse.ts
```diff
@@ -2,13 +2,23 @@ import * as t from 'io-ts';
...
+      : K
+    : never;
 }[keyof Response];
 
-export const HttpResponseCodes: { [K in Status]: number } = {
+const ensureAllResponsesDefined = <T extends { [K in Status]: number }>(
+  responseCodes: T,
+) => responseCodes;
+
+export const HttpResponseCodes = ensureAllResponsesDefined({
```
ericcrosson-bitgo, line 21: It seems like a smell to use this identity function (with penalty in code size and execution time) to enforce a type-level relationship, what about using the following patch?

```typescript
export const HttpResponseCodes = {
  ok: 200,
  invalidRequest: 400,
  unauthenticated: 401,
  permissionDenied: 403,
  notFound: 404,
  rateLimitExceeded: 429,
  internalError: 500,
  serviceUnavailable: 503,
} as const;

// Create a type-level assertion that the HttpResponseCodes map contains every key
// in the Status union of string literals, and no unexpected keys. Violations of
// this assertion will cause compile-time errors.
//
// Thanks to https://stackoverflow.com/a/67027737
type ShapeOf<T> = Record<keyof T, any>
type AssertKeysEqual<X extends ShapeOf<Y>, Y extends ShapeOf<X>> = never
type _AssertHttpStatusCodeIsDefinedForAllResponses = AssertKeysEqual<{ [K in Status]: number }, HttpResponseCodes>;
```

bitgopatmcl: Fixed in 311d8a0, but did it in a test file as an assignment check.

ericcrosson-bitgo (approved):
Looks great :slightly_smiling_face: 

:shipit:

Issue #37, opened 2022-04-09
Configure semantic-release to maintain a changelog


ericcrosson-bitgo:
We can use the following 2 semantic-release plugins

```json
[
  "@semantic-release/changelog",
  "@semantic-release/git"
]
```

but we will need to configure the changelog plugin to sign commits with gpg. There's a [section in the readme](https://github.com/semantic-release/git#gpg-signature) on this.

Because of the relative complexity that gpg signing adds, this has been cut from the initial scope of configuring semantic-release (#36)

ericcrosson-bitgo:
Moved to Jira

PR #38, opened 2022-04-09
Configure semantic-release

### LINKED ISSUE #36: Configure semantic-release
Every merge to master should invoke `semantic-release`, which will publish a package version to npm if appropriate.

Eventually we will add changelog maintenance to semantic-release via a plugin, but this has been cut from initial scope due to the complexity that our gpg-commit-signing requirement adds.

Remember that before we invoke semantic release, we need to tag our projects manually to set the initial version. Otherwise semantic-release will default to 1.0.0 for the initial version, which we would avoid if possible since we are in an alpha state.

ericcrosson-bitgo:
ci: configure semantic-release

This commit configures semantic-release to publish new versions
when semantically appropriate to npm and create a corresponding
GitHub Release.

This punts management of a changelog to #37.

Closes #36.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,40 @@
+---
+name: Release
+
+on:
+  push:
+    branches: [master, alpha, beta, next, next-major]
+
+jobs:
+  release:
+    name: Release
+    runs-one: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          cache: npm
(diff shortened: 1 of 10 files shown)


ericcrosson-bitgo:
I have added one git tag per package to [BitGo/api-ts](https://github.com/), which should set the initial version that semantic-release uses to calculate the next version numbers: https://github.com/BitGo/api-ts/tags

[Here is the documentation](https://semantic-release.gitbook.io/semantic-release/usage/configuration#existing-version-tags) on how semantic-release uses existing git tags, note that I've dropped the leading `v` as I'm following the example set by the [ericcrosson/numbers-ts](https://github.com/EricCrosson/numbers-ts/tags) repository that also uses multi-semantic-release.

bitgopatmcl (approved):

PR #41, opened 2022-04-11
fix(ci): correct typo in GitHub Action

ericcrosson-bitgo:


.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   release:
     name: Release
-    runs-one: ubuntu-latest
+    runs-on: ubuntu-latest
```
ericcrosson-bitgo, line 11: Error message [visible here](https://github.com/BitGo/api-ts/actions/runs/2151677473):

```
The workflow is not valid. .github/workflows/release.yml (Line: 11, Col: 5): Unexpected value 'runs-one' .github/workflows/release.yml (Line: 10, Col: 5): Required property is missing: runs-on
```

TheRealDwright (approved):

PR #42, opened 2022-04-12
chore: install multi-semantic-release

ericcrosson-bitgo:
In CI currently[^1] we see the following logs

```
npm WARN exec The following package was not found and will be installed: multi-semantic-release
```

which means we're not locking our CI pipeline to a specific version
of multi-semantic-release.

This commit installs a specific version of multi-semantic-release
in the root package.json.


bitgopatmcl (approved):
Approving since it looks like it does what it says. As mentioned, this isn't the cause of the current release failures so still looking into that.

PR #43, opened 2022-04-13
Fix local pkg versions

bitgopatmcl:
`multi-semantic-release`'s readme says to use `*` as the version field for monorepo packages, but this fails in a strange way. This switches them to `0.0.0-semantically-released` as recommended in a Github issue.
    
See: https://github.com/dhoulb/multi-semantic-release/issues/102

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15127,16 +15127,16 @@
     },
     "packages/express-wrapper": {
       "name": "@api-ts/express-wrapper",
-      "version": "*",
+      "version": "0.0.0-semantically-released",
       "license": "Apache-2.0",
       "dependencies": {
-        "@api-ts/io-ts-http": "*",
+        "@api-ts/io-ts-http": "0.0.0-semantically-released",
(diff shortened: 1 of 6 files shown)


bitgopatmcl:
I tried to open a PR to the branch in #42 but it wasn't working for some reason

ericcrosson-bitgo:
@bitgopatmcl counter to my message in slack -- we might as well try this again to see exactly how it blew up. I can't find the logs from the last attempt, and this won't make anything _worse_ :slightly_smiling_face: so fire in the hole

ericcrosson-bitgo (approved):
I believe I already tried this but can't find the proof of it! So I expect this to break, in a different spot than it's currently breaking, but not to make anything worse, so let's give it a shot and collect that data

ericcrosson-bitgo (approved):
Thanks for rebasing

Not sure what would have happened if we left that last commit from #42 in there but this is certainly cleaner :rocket: what a craftsman!

PR #44, opened 2022-04-18
fix: disable workspace update when running multi-semantic-release

bitgopatmcl:
See: https://docs.npmjs.com/cli/v8/using-npm/config#workspaces-update

My guess here is that the side effects of this on-by-default config option conflict with the version handling of `multi-semantic-release`. I can get `npm version` to work with this change but haven't tried it in CI yet.

Ticket: (is there one for this yet?)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,7 @@ jobs:
         run: npm test
 
       - name: Release
-        run: npx multi-semantic-release
+        run: npx --userconfig .npmrc-release multi-semantic-release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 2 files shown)


ericcrosson-bitgo:
From the linked docs: 

> If set to true, the npm cli will run an update after operations that may possibly change the workspaces installed to the node_modules folder.

Will run an _update_? What on earth does that mean :thinking: If it means `npm install`, presumably that's to update the lockfile. Would we not require a lockfile update?

Oh, no we probably wouldn't! Because the `package-lock.json` file isn't published to npm

===

Now just examining the `userconfig` file we're going to start overriding with this commit

ericcrosson-bitgo:
@bitgopatmcl do you mind pushing this branch to the BitGo/api-ts repository's `beta` branch? This should kick off a pre-release. Then we'll protect the beta branch from force-pushes

bitgopatmcl:
Okay just did that

ericcrosson-bitgo:
Thanks @bitgopatmcl, we can merge `beta` into `master` when we're ready

PR #46, opened 2022-04-18
fix: add headers to response table

bitgopatmcl:
The real motivation for this is to trigger a semantic release of
`@api-ts/response`


ericcrosson-bitgo (approved):
:ship:

PR #49, opened 2022-04-20
Use io-ts 2.1.3

bitgopatmcl:
I believe normally this would not be a patch release, but we are in pre-release beta versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -90,25 +90,25 @@
       }
     },
     "node_modules/@babel/core": {
-      "version": "7.17.8",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.17.8.tgz",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -26,6 +26,9 @@
     "test": "turbo run test"
   },
   "pre-commit": "lint-staged",
+  "overrides": {
+    "io-ts": "2.1.3"
+  },
```
ericcrosson-bitgo, line 31: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides

It is not clear to me from this documentation how `overrides` helps in this workspaces configuration

ericcrosson-bitgo (approved):
The package lock update includes transitive-dependency upgrades, but I don't think this is an issue since most of our dependencies are devDependencies.

This looks good to me, thanks for downgrading to the last, fully-stable io-ts release

PR #50, opened 2022-04-20
Increase test coverage in express wrapper

ericcrosson-bitgo:
This PR adds a test that asserts that sending a malformed request to the wrapped express server results in a 400 response with a decode error in the body.


bitgopatmcl (approved):
lgtm

PR #52, opened 2022-04-21
chore: remove extra comments

bitgopatmcl:
Removing the extra comments added in #48 now that the release caused by it has been done.


ericcrosson-bitgo (approved):
Remove a temporary workaround :+1:

Issue #53, opened 2022-04-26
Conflicting path param templates between ApiSpec expected by express-wrapper vs superagent-wrapper


jonball4:
`express-wrapper` (and `express`) expects path params denoted as `api/v1/someapi/:param/someendpoint`:

https://github.com/BitGo/api-ts/blob/5467fb41f32ada3b50f1976580305f9a6765d221/packages/express-wrapper/src/index.ts#L144

Whereas the `superagent-wrapper` expects paths with params denoted as `api/v1/someapi/{param}/someendpoint`.

https://github.com/BitGo/api-ts/blob/822eebdf28a38b64f6a19e30f53726229d814397/packages/superagent-wrapper/src/request.ts#L185

This creates the inability to satisfy both requirements / send path params through superagent-wrapper while using io-ts-http and express-wrapper.

Example in test: https://github.com/BitGo/api-ts/blob/822eebdf28a38b64f6a19e30f53726229d814397/packages/superagent-wrapper/test/request.test.ts#L14 VS. https://github.com/BitGo/api-ts/blob/822eebdf28a38b64f6a19e30f53726229d814397/packages/superagent-wrapper/test/request.test.ts#L60

ericcrosson-bitgo:
Closed by #54

PR #54, opened 2022-04-26
fix: use consistent format for path parameters between server and client

bitgopatmcl:
This is technically a breaking change, but on the other hand this was the original intended behavior.

Ticket: BG-46917

### DIFF
--- a/packages/express-wrapper/src/index.ts
+++ b/packages/express-wrapper/src/index.ts
@@ -14,6 +14,8 @@ import {
   ResponseType,
 } from '@api-ts/io-ts-http';
 
+import { apiTsPathToExpress } from './path';
+
 export type Function<R extends HttpRoute> = (
   input: RequestType<R>,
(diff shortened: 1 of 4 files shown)


bitgopatmcl:
Fixes #53

jonball4:
@bitgopatmcl why not change the path format in `superagent-wrapper` instead?  Unless...is this path format also used elsewhere?

bitgopatmcl:
It's somewhat arbitrary because the format specified by `io-ts-http` isn't meant to be tied to express. It could use that format, but would just need to be treated as coincidental.

jonball4:
> It's somewhat arbitrary because the format specified by `io-ts-http` isn't meant to be tied to express. It could use that format, but would just need to be treated as coincidental.

That makes sense, thanks.

jonball4 (approved):

ericcrosson-bitgo (approved):

PR #56, opened 2022-04-28
chore: fix inconsistent file names in io-ts-http

bitgopatmcl:


ericcrosson-bitgo:
I guess this codec doesn't have any tests :sweat_smile:

PR #59, opened 2022-05-11
feat(nix): use flakes

bitgopatmcl:


ericcrosson-bitgo (approved):
Works on my machine! I'll submit a PR with an .envrc shortly

Thanks @bitgopatmcl !

PR #62, opened 2022-05-12
chore: configure dependabot to upgrade npm packages

### LINKED ISSUE #35: Configure automatic dependency upgrades
Use something like dependabot or renovate to keep our dependencies up to date

ericcrosson-bitgo:
I'm not sure if this will work immediately or if we need to find a setting to enable dependabot for this repository. Time will tell!

I'm basing this into master because dependabot probably looks for configuration only on the default branch.

Closes #35

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  # Enable npm updates
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: daily
+      # UTC
+      time: '08:00'
+    reviewers:
+      - '@BitGo/internal-tools'
+    labels:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+version: 2
+updates:
+  # Enable npm updates
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: daily
```
ericcrosson-bitgo, line 7: Note that `daily` means "on weekdays"

https://github.com/dependabot/dependabot-core/issues/1170 :smiling_face_with_tear:

bitgopatmcl (approved):

PR #63, opened 2022-05-12
chore: use nix flakes with direnv

ericcrosson-bitgo:
Requires a fairly recent version of nix (I'm on 2.8.0) and opting in
to the following experimental features:

```
nix-command flakes
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
+.direnv/
 .drone.yml
 .nyc_output/
+.turbo/
+coverage/
 dist/
 node_modules/
-coverage/
-.turbo/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,6 +1,7 @@
+.direnv/
```
ericcrosson-bitgo, line 1: I added this line, locally I had a directory created with a few symlinks in it

.gitignore
```diff
@@ -1,6 +1,7 @@
...
 .drone.yml
 .nyc_output/
+.turbo/
+coverage/
 dist/
 node_modules/
-coverage/
-.turbo/
```
ericcrosson-bitgo, line 6: alphabetization

PR #65, opened 2022-05-12
feat(express-wrapper): allow custom response encoders

bitgopatmcl:
Fixes a soundness issue with NumericOrKeyedResponseType when instantiated with the maximally-wide HttpRoute, and is the beginning of allowing for customizing the way responses are encoded and then actually sent to Express.


ericcrosson-bitgo (approved):
Looks good to me, nice work avoiding a breaking change to the API
:rocket:

PR #76, opened 2022-05-13
ci: use lts/* to describe Node.js version

ericcrosson-bitgo:
With this indirection, we can configure a required status check that
doesn't require updating when Node.js LTS versions change.

(The matrix version appears in the name of the
status check, this was the least-gross workaround
I could find. Peer here for some horror shows:
https://github.community/t/status-check-for-a-matrix-jobs/127354/11)


ericcrosson-bitgo:
Success! We can see in the name of the status check now that we aren't describing a numeric Node.js version:

```
CI / Node.js lts/* (pull_request) 
```

.github/workflows/ci.yml
```diff
@@ -35,7 +35,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: node_modules/.cache
-          key: turbo-${{ matrix.node-version }}-${{ github.ref_name }}-${{ github.job }}-${{ github.sha }}
+          key:
+            turbo-${{ matrix.node-version }}-${{ github.ref_name }}-${{ github.job
+            }}-${{ github.sha }}
```
bitgopatmcl, line 40: Was this an auto formatting thing? This could potentially break caching.

bitgopatmcl (approved):

PR #79, opened 2022-05-16
Add router-returning function to express-wrapper

bitgopatmcl:
For existing applications that do large amounts of express
configuration, it is often cleaner to just obtain a router instance and
add it to the existing express stack. This assists with incremental
adoption of api-ts

### DIFF
--- a/packages/express-wrapper/src/index.ts
+++ b/packages/express-wrapper/src/index.ts
@@ -21,46 +21,59 @@ export type { ResponseEncoder } from './response';
 const isHttpVerb = (verb: string): verb is 'get' | 'put' | 'post' | 'delete' =>
   verb === 'get' || verb === 'put' || verb === 'post' || verb === 'delete';
 
-export const createServerWithResponseEncoder =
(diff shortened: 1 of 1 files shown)


packages/express-wrapper/src/index.ts
```diff
@@ -21,46 +21,59 @@ export type { ResponseEncoder } from './response';
...
+  encoder?: ResponseEncoder;
+};
 
-        const expressPath = apiTsPathToExpress(httpRoute.path);
-        router[method](expressPath, handlers);
+export function routerForApiSpec<Spec extends ApiSpec>({
+  spec,
+  routeHandlers,
+  encoder = defaultResponseEncoder,
+}: CreateRouterProps<Spec>) {
```
ericcrosson-bitgo, line 38: Interesting choice of parameter type! (an object)

ericcrosson-bitgo (approved):
Would love to see some docstrings explaining some high-level concepts; for example, when to choose routerForApiSpec vs createServer.

PR #80, opened 2022-05-16
Merge master into beta

ericcrosson-bitgo:
This PR merges `master` into `beta`, bringing the commits from PRs landed directly into `master` into `beta`.

Specifically, we want to bring in 80572985356e88b2c0e2cf79ac7a5227300b2bde which modifies the GitHub Actions stage name, so we can enable required status checks for the latest lts.

Ticket: BG-44133


bitgopatmcl:
Looks like we got an `io-ts-types` version mismatch (hopefully it is that simple): https://github.com/BitGo/api-ts/runs/6456427033?check_suite_focus=true#step:7:77

ericcrosson-bitgo:
> Looks like we got an `io-ts-types` version mismatch (hopefully it is that simple): https://github.com/BitGo/api-ts/runs/6456427033?check_suite_focus=true#step:7:77

Thanks for checking! That looks about right... but I have no idea how :sweat_smile: I'll mess with this locally

PR #84, opened 2022-05-17
chore: remove explicit peer dependencies

ericcrosson-bitgo:
We currently list monocle-ts and newtype-ts as explicit dependencies,
since these two packages are peer dependencies of io-ts-types[^1]
and npm 6 doesn't automatically install peer dependencies (like npm
7 does).

This commit removes the explicit peer dependencies, because when npm 7+
is used the behavior will be the same. Additionally, in npm 6 these
to packages won't be installed but they aren't needed, so there will
be no breaking change.

[^1]: https://github.com/gcanti/io-ts-types/blob/master/package.json#L40-L41

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5027,6 +5027,7 @@
     "node_modules/monocle-ts": {
       "version": "2.3.10",
       "license": "MIT",
+      "peer": true,
       "peerDependencies": {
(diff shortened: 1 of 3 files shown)


bitgopatmcl (approved):
lgtm

PR #94, opened 2022-05-20
refactor(express-wrapper): avoid using codec type-guard

ericcrosson-bitgo:
This refactor changes the response code path to try encoding the
service function's response value with the apiSpec's response codec,
and if this errors returns a 500.

Previously we used the codec's type-guard as a test if the codec
can encode the service function's response value, but this step will
take extra time and there's no hard contract that guarantees us the
type-guard returning `true` really means that the codec will be able
to encode a value. Instead, we just try the dangerous deed and do our
best to catch errors.


packages/express-wrapper/src/response.ts
```diff
@@ -38,13 +38,12 @@ export const defaultResponseEncoder: ResponseEncoder = (
...
+  } catch (error) {
     console.warn(
       "Unable to encode route's return value, did you return the expected type?",
     );
     expressRes.status(500).end();
-    return;
   }
-
-  expressRes.status(status).json(responseCodec.encode(payload)).end();
 };
```
ericcrosson-bitgo, line 49: Proving that `responseCodec` isn't undefined is pretty gross in this implementation :sweat: I used a `!` because `undefined has no  encode method` will still jump to the `catch` block

bitgopatmcl (approved):
Thanks!

PR #105, opened 2022-05-26
feat: make superagent/test a peer dependency

bitgopatmcl:
superagent-wrapper works with these packages, but shouldn't be pulling
them in itself.


ericcrosson-bitgo (commented):
Do you think we should hold off on this until Node.js versions bundled with npm 6 are no longer LTS versions?

This will be a breaking change for Node.js 14, since superagent will no longer be installed

PR #106, opened 2022-05-26
fix(deps): @types packages are necessary for types inferencing

ericcrosson-bitgo:
so add @types/superagent and @types/supertest to dependencies so that
downstream users get the expected type-inferencing out of the box


bitgopatmcl (approved):
Adds missing `@types` packages

PR #108, opened 2022-05-31
Use prettier in ci

ericcrosson-bitgo:


bitgopatmcl (approved):
Testted locally and the CI config looks good

PR #120, opened 2022-06-03
chore: remove go from flake.nix

bitgopatmcl:
Not sure how this snuck in there, but it is not needed to build anything


ericcrosson-bitgo (approved):
Remove Go from flake.nix

PR #122, opened 2022-06-03
fix(superagent): do not throw on non-2xx status

bitgopatmcl:
Ticket: BG-47455

### DIFF
--- a/packages/superagent-wrapper/test/request.test.ts
+++ b/packages/superagent-wrapper/test/request.test.ts
@@ -5,9 +5,11 @@ import express from 'express';
 import * as E from 'fp-ts/Either';
 import * as t from 'io-ts';
 import { NumberFromString } from 'io-ts-types';
+import superagent from 'superagent';
 import supertest from 'supertest';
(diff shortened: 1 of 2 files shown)


packages/superagent-wrapper/test/request.test.ts
```diff
@@ -184,4 +186,32 @@ describe('request', () => {
...
+        TestRoutes,
+      );
+
+      const req = await superagentClient['api.v1.test']
+        .post({ id: 1337, foo: 'test', bar: 42 })
+        .set('x-send-unexpected-status-code', 'true')
+        .decode();
+
+      assert.equal(req.status, 'decodeError');
+      assert.equal(req.original.status, 400);
```
ericcrosson-bitgo, line 214: Ah, so here's how we get the value behind the non-standard 'decodeError'. I still feel weird about having numeric status codes _except_ for 'decodeError' but still don't have a great improvement in mind

ericcrosson-bitgo (approved):

PR #127, opened 2022-06-21
fix API Reference so that is correctly links to GitHub from NPM

andrew-scott-fischer:


ericcrosson-bitgo:
Fastest merge in the west :cowboy_hat_face:

PR #130, opened 2022-06-23
Structural superagent dependency

bitgopatmcl:
`@api-ts/superagent-wrapper` only actually uses a small subset of superagent's functionality. This PR adds the necessary type declarations so that it can be compatible with the widest possible range of superagent versions.

### DIFF
--- a/packages/superagent-wrapper/src/request.ts
+++ b/packages/superagent-wrapper/src/request.ts
@@ -1,8 +1,6 @@
 import * as h from '@api-ts/io-ts-http';
 import * as E from 'fp-ts/Either';
 import * as t from 'io-ts';
-import type { Response, SuperAgent, SuperAgentRequest } from 'superagent';
-import type { SuperTest } from 'supertest';
 import { URL } from 'url';
 import { pipe } from 'fp-ts/function';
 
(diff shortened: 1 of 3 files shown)


packages/io-ts-http/src/httpRoute.ts
```diff
@@ -23,8 +23,8 @@ export type ResponseType<T extends HttpRoute> = {
 }[keyof T['response']];
 
 export type ApiSpec = {
-  [Key: string]: {
-    [Req: string]: HttpRoute;
+  readonly [Key: string]: {
+    readonly [Req: string]: HttpRoute;
```
bitgopatmcl, line 27: oops, this snuck in, reverting

ericcrosson-bitgo (approved):

packages/superagent-wrapper/src/request.ts
```diff
@@ -40,6 +38,22 @@ type PatchedRequest<Req extends SuperAgentRequest, Route extends h.HttpRoute> =
...
+  body: unknown;
+  status: number;
+};
+
+interface SuperagentRequest<Res extends Response> extends Promise<Res> {
+  ok(callback: (response: Res) => boolean): this;
+  query(params: Record<string, string | string[]>): this;
+  set(name: string, value: string): this;
+  send(body: string): this;
+}
```
ericcrosson-bitgo, line 55: Does this implementation mean we won't learn if superagent changes out from underneath us?

@bitgopatmcl

PR #131, opened 2022-06-23
Update superagent

bitgopatmcl:
Now that superagent is just a dev dependency, might as well just bump it to the newest version.

Obsoletes #113


ericcrosson-bitgo (approved):
Love me the confidence of types

PR #140, opened 2022-07-01
Upgrade TypeScript to 4.7.4

bitgopatmcl:
This led to quite a bit of refactoring of type definitions, but I believe they are more solid now.

### DIFF
--- a/packages/io-ts-http/src/combinators.ts
+++ b/packages/io-ts-http/src/combinators.ts
@@ -1,12 +1,21 @@
+import { pipe } from 'fp-ts/function';
 import * as E from 'fp-ts/Either';
 import * as R from 'fp-ts/Record';
 import * as t from 'io-ts';
-import { Flattened, FlattenedC, NestedC, PossiblyUndefinedKeys } from './utils';
+import {
+  Flattened,
+  NestedType,
+  NestedOutputType,
+  NestedProps,
+  OptionalizedC,
+  OptionalProps,
+  RequiredProps,
+} from './utils';
 
 export const optional = <C extends t.Mixed>(subCodec: C) =>
   t.union([subCodec, t.undefined]);
 
-export const optionalized = <P extends t.Props>(props: P) => {
+export const optionalized = <P extends t.Props>(props: P): OptionalizedC<P> => {
   const requiredProps: t.Props = {};
   const optionalProps: t.Props = {};
   for (const key in props) {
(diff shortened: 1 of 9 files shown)


packages/io-ts-http/src/combinators.ts
```diff
@@ -22,46 +31,49 @@ export const optionalized = <P extends t.Props>(props: P) => {
...
     props,
   );
-  const nestedCodec = t.strict(nestedProps) as unknown as NestedC<Props>;
+  const nestedCodec = t.strict(nestedProps);
 
-  return new t.Type<t.TypeOf<FlattenedC<Props>>, t.OutputOf<NestedC<Props>>>(
+  return new t.Type(
     name,
-    flatCodec.is,
+    flatCodec.is as any,
```
ericcrosson-bitgo, line 60: eslint-fu not strong enough yet! :grin:

packages/io-ts-http/src/combinators.ts
```diff
@@ -22,46 +31,49 @@ export const optionalized = <P extends t.Props>(props: P) => {
...
     (input, context) =>
-      E.map((nested: t.TypeOf<NestedC<Props>>) => {
-        let flattened: Partial<t.TypeOf<FlattenedC<Props>>> = {};
-        for (const key in nested) {
-          if (!nested.hasOwnProperty(key)) {
-            continue;
+      pipe(
+        nestedCodec.validate(input, context),
+        E.map((nested) => {
+          let flattened = {};
```
ericcrosson-bitgo, line 65: I prefer the before diff here, where we had a manual type declaration for this value.
Otherwise {} means "any non-nullish value" [source](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/ban-types.md#default-options) and I'm not sure if the case on line 72 protects us against assignability fails

packages/io-ts-http/src/httpRequest.ts
```diff
@@ -18,20 +18,16 @@ export type HttpRequestCodec<T> = t.Type<
 >;
 
 export type HttpRequestCombinatorProps = {
-  params?: OutputConstrainedProps<string | undefined>;
-  query?: OutputConstrainedProps<string | string[] | undefined>;
-} & FieldPowerSet<{
```
ericcrosson-bitgo, line 23: No more FieldPowerSet!

ericcrosson-bitgo (approved):
Agreed, this is large but I'm in favor of many changes in this PR. Seems a net reduction in complexity. I do see some lines where my primitive human brain can't act well enough as a compiler to see if we have preserved the same level of guarantee, but these appear to be contained within functions with well-typed signatures, so I'm presuming they are okay

Issue #138, opened 2022-06-30
using PathReporter for showing decoding errors leads to thrown errors when using bigint


jonball4:
https://github.com/BitGo/api-ts/blob/2c1e3900d30a9fa5f1d273fcf8b16933044aa091/packages/express-wrapper/src/request.ts#L52-L55

https://github.com/gcanti/io-ts/blob/920b21ea32e82b328cd14ff4b28da72ae60db505/src/PathReporter.ts#L8-L29

Since PathReporter uses JSON.stringify, which does not have handling for `bigint` by default, this causes an error to get thrown. 
 The result is that the request never gets a response and times out.
 
Suggesting we pull in our internal `validationErrors` routine from schema-lib, or our own version of the PathReporter.

ericcrosson-bitgo:
> The result is that the request never gets a response and times out.

Yikes, we'll fix this. Thanks for the report!

Can you point me to the codec with a bigint? I'd like to see more about the input that causes this behavior

jonball4:
sure!  Here it is:

https://github.com/BitGo/bitgo-microservices/blob/056251f23a001d0cf9549f49efc1a4b7380034c6/packages/bitgo-network/src/controllers/v1/allocation/response.ts#L5-L20

bitgopatmcl:
Fixed in #267

jonball4:
> Fixed in #267

almost..  PathReport will

jonball4:
I suppose this is not an api-ts issue anymore.

Issue #137, opened 2022-06-29
superagent-wrapper's decodeExpecting does not decode error responses


jonball4:
The experience using superagent-wrapper in production code would be much improved if `decodeExpecting` would also decode the response if the expected status code was not returned by the request.  

In the current beta version, a barebones error message is returned complaining about an incorrect status code:

https://github.com/BitGo/api-ts/blob/8a460d6d629eb7cdc84afb2b195347242e9cf8d4/packages/superagent-wrapper/src/request.ts#L153-L167

The error message would be more helpful if it included a stringified version of the decoded error response, if a codec is defined for the unexpected status code returned.

Some food for thought - In order to work around this limitation of `decodeExpecting`, I had written a helper method in one of my use cases to create a more useful error message taking the error response into account:

```typescript
const unknownResponseToError = (operation: string) => (response: { body: unknown }) => pipe(
  response.body,
  Json.stringify,
  E.match(
    () => new BNError('DAS API Request returned an unreadable response', DAS_API_ERROR, response),
    (stringifiedBody: string) => new BNError(
      `DAS API Request for '${operation}' returned an unexpected response: ${stringifiedBody}`,
      DAS_API_ERROR
    ),
  )
);

const processDecodedResponse = <R extends HttpRoute>(
  httpRoute: R, operation: string,
) => TE.chainEitherK(
    (decodedResponse: {
      body: unknown
  }): E.Either<Error, t.TypeOf<R['response'][200]>> => pipe(
      decodedResponse.body,
      E.fromPredicate(
        httpRoute.response[200].is,
        () => unknownResponseToError(operation)(decodedResponse)
      )
    )
  );

```

bitgopatmcl:
It should be possible to do this, although `decodeExpecting` throws when it gets the unknown status code, so the catching code would need to handle the wider error type.

Does the regular `decode` work for your use-case? It isn't supposed to throw (it used to because of unexpected behavior in superagent, but that has been fixed), and it gives you a tagged union of all the possible responses along with a catch-all one for decoding errors.

jonball4:
> > If `decodeExpecting` threw an error including the body of the response, then I would be able wrap it in a `TE.tryCatch` and just pass along the error via `identity`
> 
> Do you have an assumption that the `thrown` value will always be an `Error`? I'm trying to make sense of this last sentence in the context of the preceding text

Yes

bitgopatmcl:
[...]

PR #162, opened 2022-07-08
feat: add decoded response to decodeExpecting errors

bitgopatmcl:
Should address #137 

Also it brings in `PathReporter` which has a bug as per #138, but we were calling `JSON.stringify` on the error anyway so it doesn't introduce a regression. Will try to sort that out in a follow-up.

### DIFF
--- a/packages/superagent-wrapper/src/request.ts
+++ b/packages/superagent-wrapper/src/request.ts
@@ -1,6 +1,7 @@
 import * as h from '@api-ts/io-ts-http';
 import * as E from 'fp-ts/Either';
 import * as t from 'io-ts';
+import * as PathReporter from 'io-ts/lib/PathReporter';
 import { URL } from 'url';
 import { pipe } from 'fp-ts/function';
 
@@ -13,14 +14,23 @@ type SuccessfulResponses<Route extends h.HttpRoute> = {
   };
 }[keyof Route['response']];
 
-type DecodedResponse<Route extends h.HttpRoute> =
+export type DecodedResponse<Route extends h.HttpRoute> =
   | SuccessfulResponses<Route>
   | {
       status: 'decodeError';
-      error: unknown;
+      error: string;
       body: unknown;
       original: Response;
     };
+
+export class DecodeError extends Error {
+  readonly decodedResponse: DecodedResponse<h.HttpRoute>;
+
+  constructor(message: string, decodedResponse: DecodedResponse<h.HttpRoute>) {
+    super(message);
+    this.decodedResponse = decodedResponse;
+  }
+}
 
 const decodedResponse = <Route extends h.HttpRoute>(res: DecodedResponse<Route>) => res;
 
@@ -134,7 +144,7 @@ const patchRequest = <
           // DISCUSS: what's this non-standard HTTP status code?
           decodedResponse<Route>({
             status: 'decodeError',
-            error,
+            error: PathReporter.failure(error).join('\n'),
             body: res.body,
             original: res,
           }),
@@ -146,9 +156,16 @@ const patchRequest = <
     status: StatusCode,
   ) =>
     patchedReq.decode().then((res) => {
-      if (res.status !== status) {
-        const error = res.error ?? `Unexpected status code ${String(res.status)}`;
-        throw new Error(JSON.stringify(error));
+      if (res.original.status !== status) {
+        const error = `Unexpected response ${String(
+          res.original.status,
+        )}: ${JSON.stringify(res.original.body)}`;
+        throw new DecodeError(error, res as DecodedResponse<h.HttpRoute>);
+      } else if (res.status === 'decodeError') {
+        const error = `Could not decode response ${String(
+          res.original.status,
+        )}: ${JSON.stringify(res.original.body)}`;
+        throw new DecodeError(error, res as DecodedResponse<h.HttpRoute>);
       } else {
         return res as ExpectedDecodedResponse<Route, StatusCode>;
       }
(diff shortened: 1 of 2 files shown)


packages/superagent-wrapper/src/request.ts
```diff
@@ -148,7 +158,7 @@ const patchRequest = <
     patchedReq.decode().then((res) => {
       if (res.status !== status) {
         const error = res.error ?? `Unexpected status code ${String(res.status)}`;
```
ericcrosson-bitgo, line 160: `error` perhaps a misleading name for this variable now

ericcrosson-bitgo (approved):
Looks fine to me, I don't see any breaking changes.

I quite dislike `Error`s and especially subclassing `Error`s, but `decodeExpecting` is designed to `throw` and there's no better value to `throw` than an `Error`.

Lastly, I want to make sure this change identifies the entirety of #137. I see

> The error message would be more helpful if it included a stringified version of the decoded error response

as best I can tell this is the main ask, so looks like this PR hits the :dart:

ericcrosson-bitgo:
Oh, I think I see what you mean. So instead of saying "I tried to decode 200Codec and got these errors", it's sufficient to say "I expected a 200 but got this status code and response instead"?

bitgopatmcl:
Maybe it would be helpful to list out the possibilities and what should happen:

- Expected status code, decoded correctly: nothing thrown, result returned like normal
- Unexpected status code, decoded correctly: error thrown: "Unexpected response 404: { message: "Thing not found" }"
- Expected status code, decode errors: error thrown: "Could not decode response 200: { stuff: [ ... ] }"
- Unexpected status code, decode errors: error thrown: "Unexpected response 418: { message: "I'm a teapot" }"
- Unknown status code: error thrown: "Unexpected response 404: { error: "oops" }"

Do these seem right?

packages/superagent-wrapper/test/request.test.ts
```diff
@@ -169,21 +181,32 @@ describe('request', () => {
...
         .post({ id: 1337, foo: 'test', bar: 42 })
         .set('x-send-invalid-response-body', 'true')
         .decodeExpecting(200)
-        .then(() => false)
-        .catch(() => true);
+        .then(() => '')
+        .catch((err) => (err instanceof DecodeError ? err.message : ''));
 
-      assert.isTrue(result);
+      assert.deepEqual(result, 'Could not decode response 200: {"invalid":"response"}');
```
ericcrosson-bitgo, line 209: `deepEqual` an odd choice here!

packages/superagent-wrapper/src/request.ts
```diff
@@ -146,9 +156,16 @@ const patchRequest = <
...
-        throw new Error(JSON.stringify(error));
+      if (res.original.status !== status) {
+        const error = `Unexpected response ${String(
+          res.original.status,
+        )}: ${JSON.stringify(res.original.body)}`;
+        throw new DecodeError(error, res as DecodedResponse<h.HttpRoute>);
+      } else if (res.status === 'decodeError') {
+        const error = `Could not decode response ${String(
+          res.original.status,
+        )}: ${JSON.stringify(res.original.body)}`;
```
ericcrosson-bitgo, line 167: Note that `JSON.stringify` can also throw :sob: I believe when a circular reference is encountered.

fp-ts [provides a wrapper](https://gcanti.github.io/fp-ts/modules/Json.ts.html#stringify) that returns an `Either`, which we may want to use

packages/superagent-wrapper/src/request.ts
```diff
@@ -146,9 +156,16 @@ const patchRequest = <
     status: StatusCode,
   ) =>
     patchedReq.decode().then((res) => {
-      if (res.status !== status) {
-        const error = res.error ?? `Unexpected status code ${String(res.status)}`;
-        throw new Error(JSON.stringify(error));
+      if (res.original.status !== status) {
+        const error = `Unexpected response ${String(
+          res.original.status,
+        )}: ${JSON.stringify(res.original.body)}`;
```
ericcrosson-bitgo, line 162: Is `res.original.status` not already coerced to a string by the backticks?
```suggestion
        const error = `Unexpected response ${res.original.status}: ${JSON.stringify(res.original.body)}`;
```

ericcrosson-bitgo:
> Maybe it would be helpful to list out the possibilities and what should happen:

I hope we have tests for each case, I'll take a look

ericcrosson-bitgo (approved):
This change looks fine to me, it modifies `decodeExpecting` to return a particular error (`DecodeError` -- note @bitgopatmcl that this may be ambiguous with an io-ts decode error _outside_ of an api-ts context) so we can identify this case with `instanceof`. 

Additionally, change the error string in the `message` of the `DecodeError` to be more useful as determined by our users. 

I dig it

packages/superagent-wrapper/src/request.ts
ericcrosson-bitgo: https://github.com/BitGo/api-ts/pull/170

packages/superagent-wrapper/test/request.test.ts
ericcrosson-bitgo: https://github.com/BitGo/api-ts/pull/171

packages/superagent-wrapper/src/request.ts
ericcrosson-bitgo: hurm how would you want to handle this case, @bitgopatmcl? It's pretty gnarly

I was about to open a ticket to track and then I had the thought -- is it even possible for `res.original.body` to contain a circular reference here? Meaning, was `res.original.body` recently `JSON.parse`d? If so, then I suppose there's no cause for concern (though I'll leave a note indicating that we've considered the unsafe use of `JSON.stringify` and determined that we're in the clear)

bitgopatmcl: What about if the server returns a non-`application/json` response? We should probably consider that case too.

jonball4:
🍾  thank you guys

PR #167, opened 2022-07-19
feat: force simplification of intersection types in io-ts-http

bitgopatmcl:
Adds a utility type that forces the TS language server to simplify
intersection types (at least as of 4.7.4). This greatly improves
readability of the codec types produced by io-ts-http

Before:
image

After:
image


ericcrosson-bitgo (approved):
Brilliant! Thanks for the pics of the LSP overlays

PR #168, opened 2022-07-19
feat: produce helpful error messages on httpRequest codec errors

bitgopatmcl:
Attempts to produce a helpful error message when invalid codecs are passed to `httpRequest`.
It is a workaround inspired by [this comment](https://github.com/microsoft/TypeScript/pull/40468#issuecomment-1162386217)

Before:
image

After:
image

In this case the "before" error isn't too bad, but this technique could be useful in other places.


ericcrosson-bitgo (approved):
Nice improvement! I didn't know such a formatting was possible. 

@louib-bitgo I'm not sure I follow, are you able to push a quick proof of concept to make sure we're on the same page?

PR #172, opened 2022-07-20
chore(deps): use c8 instead of nyc

ericcrosson-bitgo:
This commit moves us one step closer to consistency among our api-ts
packages in this monorepo. c8 is desirable over nyc, though they're
similar in functionality, because it leverages a built-in Node.js
feature for collecting code coverage instead of annotating test files
before test execution. Consequently, test runtimes are faster and
reporting is more accurate.


louib-bitgo (approved):
There are still a few nyc leftovers, like in `.gitignore` and `.prettierignore`

ericcrosson-bitgo:
> There are still a few nyc leftovers, like in `.gitignore` and `.prettierignore`

@louib-bitgo That's easy enough to fix in a follow-up PR :smiley: 

https://github.com/BitGo/api-ts/pull/174

PR #177, opened 2022-07-25
feat(express-wrapper): allow middleware to add request properties

### LINKED ISSUE #134: Express-wrapper throws out middleware data from the request unless included in request codec
In the current beta version, if we define route middleware via `express-wrapper` as such:

```
const exampleMiddleware = (apiName: string, method: 'get' | 'post' | 'put' | 'delete') => (req, res, next) => {
  req.apiName = apiName;
  req.method = method;
  next();
}

const createRouteHandler = <A, B>(
  apiName: string,
  method: 'get' | 'post' | 'put' | 'delete',
  handler: (_: A) => B,
) => ({
    middleware: [exampleMiddleware(apiName, method)],
    handler,
  })

export const v1ExampleRoute = httpRoute({
  path: 'my/example/path',
  method: 'GET',
  request: httpRequest({ parameter: t.string }),
  response: {
    200: t.type({ result: t.string }),
  },
});

export const v1ExampleSpec = apiSpec({
  ['v1.example.route']: {
    get: v1ExampleRoute
  },
});

const app = createServer(v1ExampleSpec, (expressApp) => (
  {
    ['v1.example.route']: {
      get: createRouteHandler('v1.example.route', 'get', handleExampleRequest),
    }
));
```

The `express-wrapper` uses the `httpRoute.request` codec to decode the express request and only passes the decoded values.  Anything else the middleware up til that request handler might have added to the request will be thrown out, unless the request codec includes the values added by the middleware.

https://github.com/BitGo/api-ts/blob/7dc00a4500104ff0b78b9f81bfcb68037dfe5a12/packages/express-wrapper/src/request.ts#L49

I think this warrants the need for a separate _middleware codec_ defined at the express server configuration level, to process the request w.r.t. the server level middleware in use.   Additionally, route-specific middleware warrants the need for route-specific middleware codecs as well.

bitgopatmcl:
This adds the capability for middleware to define additional properties that get added on to the decoded request object. This is done with the addition of a new exported helper function named `middlewareFn`. It accepts a function with the signature:

```
(req: express.Request, res: express.Response) => Promise<Output>
```

Notice that it is `async` rather than using Express's `next` convention. Properties of `Output` are merged in with the decoded request.

This change should be backwards compatible with existing use of `express-wrapper`. Middleware functions just go in the pipeline alongside normal express middleware. However, it introduces a middleware array length constraint of 8. This is just how much I wanted to write out for now, it can be expanded or we can try to figure out recursive tuple expansion in TS if it needs to be longer.

Hopefully resolves #134 

TODO:
- [x] Tests
- [x] Document `middlewareFn`

### DIFF
--- a/packages/express-wrapper/src/middleware.ts
+++ b/packages/express-wrapper/src/middleware.ts
@@ -1,0 +1,206 @@
+import * as express from 'express';
+
+export const MiddlewareBrand = Symbol();
+export type MiddlewareBrand = typeof MiddlewareBrand;
+
+export type MiddlewareFn<T extends {}> = (
+  req: express.Request,
+  res: express.Response,
+) => Promise<T>;
+
+type MiddlewareHandler<T extends {} = {}> = {
+  (req: express.Request, res: express.Response, next: express.NextFunction): void;
+  [MiddlewareBrand]: T;
+};
+
+export type MiddlewareRequestHandler = express.RequestHandler | MiddlewareHandler;
+
+type MiddlewareOutput<T extends MiddlewareRequestHandler> = T extends {
+  [MiddlewareBrand]: infer R;
+}
+  ? R
+  : {};
+
+type MiddlewareNextFn = (error?: unknown, middlewareResult?: {}) => void;
+
+/**
+ * Creates an express request handler that also adds properties to the incoming decoded request. Any existing properties
+ * with conflicting names will be replaced.
+ *
+ * @param fn - an async function that reads the express req/res and returns the desired additional properties.
+ * @returns
+ */
+export function middlewareFn<T extends {}>(fn: MiddlewareFn<T>): MiddlewareHandler<T> {
+  const result: express.RequestHandler = (req, res, next: MiddlewareNextFn) => {
+    fn(req, res)
+      .then((result) => next(undefined, result))
+      .catch((err) => next(err));
+  };
+  return result as MiddlewareHandler<T>;
+}
+
+type MiddlewareResult<Input extends {}, T> = T extends MiddlewareRequestHandler
+  ? Omit<Input, keyof MiddlewareOutput<T>> & MiddlewareOutput<T>
+  : never;
+
+export type MiddlewareChain =
+  | []
+  | [MiddlewareRequestHandler]
+  | [MiddlewareRequestHandler, MiddlewareRequestHandler]
+  | [MiddlewareRequestHandler, MiddlewareRequestHandler, MiddlewareRequestHandler]
+  | [
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+    ]
+  | [
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+    ]
+  | [
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+    ]
+  | [
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+    ];
+
+export type MiddlewareChainOutput<
+  Input,
+  Chain extends MiddlewareChain,
+> = Chain extends []
+  ? Input
+  : Chain extends [infer A]
+  ? MiddlewareResult<Input, A>
+  : Chain extends [infer A, infer B]
+  ? MiddlewareResult<MiddlewareResult<Input, A>, B>
+  : Chain extends [infer A, infer B, infer C]
+  ? MiddlewareResult<MiddlewareResult<MiddlewareResult<Input, A>, B>, C>
+  : Chain extends [infer A, infer B, infer C, infer D]
+  ? MiddlewareResult<
+      MiddlewareResult<MiddlewareResult<MiddlewareResult<Input, A>, B>, C>,
+      D
+    >
+  : Chain extends [infer A, infer B, infer C, infer D, infer E]
+  ? MiddlewareResult<
+      MiddlewareResult<
+        MiddlewareResult<MiddlewareResult<MiddlewareResult<Input, A>, B>, C>,
+        D
+      >,
+      E
+    >
+  : Chain extends [infer A, infer B, infer C, infer D, infer E, infer F]
+  ? MiddlewareResult<
+      MiddlewareResult<
+        MiddlewareResult<
+          MiddlewareResult<MiddlewareResult<MiddlewareResult<Input, A>, B>, C>,
+          D
+        >,
+        E
+      >,
+      F
+    >
+  : Chain extends [infer A, infer B, infer C, infer D, infer E, infer F, infer G]
+  ? MiddlewareResult<
+      MiddlewareResult<
+        MiddlewareResult<
+          MiddlewareResult<
+            MiddlewareResult<MiddlewareResult<MiddlewareResult<Input, A>, B>, C>,
+            D
+          >,
+          E
+        >,
+        F
+      >,
+      G
+    >
+  : never;
+
+/**
+ * Runs a middleware chain, and adds any properties returned by middleware..
+ *
+ * @param input - the decoded request properties
+ * @param chain - the middleware chain
+ * @param req - express request object
+ * @param res - express response object
+ * @returns `input` with possible additional properties as specified by the middleware chain
+ */
+export async function runMiddlewareChain<
+  Input extends {},
+  Chain extends MiddlewareChain,
+>(
+  input: Input,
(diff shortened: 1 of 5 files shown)


ericcrosson-bitgo (commented):
Is it possible to write tests for this diff? We should strive for 100% test coverage in this library, or barring that, monotonically-increasing test coverage

packages/express-wrapper/src/request.ts
```diff
@@ -72,3 +86,32 @@ export const decodeRequestAndEncodeResponse = (
...
+    const nextValueOrErr: unknown = await new Promise((resolve, reject) => {
+      try {
+        middleware(req, res, resolve);
+      } catch (err) {
+        reject(err);
+      }
+    });
+
+    let newResult = extractOutput(nextValueOrErr);
+    if (!newResult) {
```
ericcrosson-bitgo, line 110: Can we explicitly check for `undefined` here? To match the value returned in the `else` in `extractOutput`
```suggestion
    if (newResult === undefined) {
```

packages/express-wrapper/src/middleware.ts
```diff
@@ -0,0 +1,130 @@
...
+          E
+        >,
+        F
+      >,
+      G
+    >
+  : never;
+
+export function extractOutput(value: unknown): {} | undefined {
+  if (value && typeof value === 'object' && MiddlewareBrand in value) {
```
ericcrosson-bitgo, line 125: I'll be a human-operator for [eslint/strict-boolean-test](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md) until we have eslint wired up :smiley: 

I think this PR in particular would see increased readability of intent with explicit boolean tests

ericcrosson-bitgo: On second thought -- can't we omit `value &&` entirely in this expression?

packages/express-wrapper/src/request.ts
```diff
@@ -72,3 +86,32 @@ export const decodeRequestAndEncodeResponse = (
...
+export async function runMiddlewareChain<
+  Input extends {},
+  Chain extends MiddlewareChain,
+>(
+  input: Input,
+  chain: Chain,
+  req: express.Request,
+  res: express.Response,
+): Promise<MiddlewareChainOutput<Input, Chain>> {
+  let result: {} = input;
```
ericcrosson-bitgo, line 99: I tried to make this clearer by writing
```typescript
  let result: Input = input;
```

but after making the other necessary edits to appease the type-checker I think I confused myself more :sweat_smile: 
This is some classic value-focused JS programming!

ericcrosson-bitgo (approved):
Looks fine to me, comments are minor enough to address in subsequent PRs

bitgopatmcl:
> Is it possible to write tests for this diff?

Yup, they won't take very long to write, just wanted to get this draft up to solicit feedback.

packages/express-wrapper/src/middleware.ts
jonball4: `value` can be `null` and also have a `typeof value === 'object'`?

packages/express-wrapper/src/request.ts
```diff
@@ -8,15 +8,22 @@ import * as PathReporter from 'io-ts/lib/PathReporter';
...
-export type ServiceFunction<R extends HttpRoute> = (
-  input: RequestType<R>,
+export type ServiceFunction<R extends HttpRoute, Input = RequestType<R>> = (
+  input: Input,
 ) => NumericOrKeyedResponseType<R> | Promise<NumericOrKeyedResponseType<R>>;
 
-export type RouteHandler<R extends HttpRoute> =
+export type RouteHandler<
+  R extends HttpRoute,
+  Chain extends MiddlewareChain = MiddlewareChain,
```
jonball4, line 20: just curious, why do we need a default generic type here?

packages/express-wrapper/src/middleware.ts
```diff
@@ -0,0 +1,130 @@
...
+      MiddlewareRequestHandler,
+      MiddlewareRequestHandler,
+    ];
+
+export type MiddlewareChainOutput<
+  Input,
+  Chain extends MiddlewareChain,
+> = Chain extends []
+  ? Input
+  : Chain extends [infer A]
```
jonball4, line 78: what is the effect of `infer` here?

bitgopatmcl: it assigns the subtype of the first (only in this case) tuple element to A. Here it will either be a normal express request handler, or one of the ones wrapped by `middlewareFn` above. In the latter case it is then possible to figure out what properties that middleware provides so they can be added to the request type.

packages/express-wrapper/src/middleware.ts
ericcrosson-bitgo: `null`  _is_ an object!

```
; node
Welcome to Node.js v14.19.2.
Type ".help" for more information.
> typeof null
'object'
> 
```

jonball4: thats fun, thanks for clearing that up :)

packages/express-wrapper/src/request.ts
bitgopatmcl: Need a way to relate the type produced by the middleware to the input of the ServiceFunction. There might be another way to do this though.

packages/express-wrapper/src/middleware.ts
jonball4: got it, hadn't seen this used in practice til now, thanks! took on some light reading last night..makes sense to me now

packages/express-wrapper/src/request.ts
bitgopatmcl: Yeah discovered while writing tests that something isn't quite right here. The type inferencing works out for the middleware stuff, but when I go to combine it with the whole api spec then it just gets inferred as the widest possible type.

packages/express-wrapper/src/middleware.ts
```diff
@@ -0,0 +1,130 @@
+import * as express from 'express';
+
+const MiddlewareBrand = Symbol();
+type MiddlewareBrand = typeof MiddlewareBrand;
+
+export type MiddlewareFn<T extends {}> = (
+  req: express.Request,
+  res: express.Response,
+) => Promise<T>;
```
bitgopatmcl, line 9: Realized there is a potential safety issue here because of how strict/exact types can't really be enforced without something like io-ts doing it at runtime. A middleware can just clobber request properties with arbitrary values by just adding them on as extra values and claiming to return a wider type (or even `{}`).

bitgopatmcl: Planning on fixing it by just adding a codec param to `middlewareFn` that requires an exact codec for the desired properties.

bitgopatmcl: actually wait, that just moves the problem elsewhere, but might be worth doing still since it is less likely for someone to accidentally add properties to a codec definition versus a function return.

image

packages/express-wrapper/test/server.test.ts
```diff
@@ -139,7 +140,7 @@ test('should handle io-ts-http formatted path parameters', async (t) => {
     app.use(appMiddleware);
     return {
       'hello.world': {
-        put: { middleware: [routeMiddleware], handler: CreateHelloWorld },
+        put: routeHandler([routeMiddleware], CreateHelloWorld),
```
ericcrosson-bitgo, line 143: One nit: we've been using the term "route handler" to refer to the function that ultimately fields the http request (in this case, `CreateHelloWorld`). The introduced `routeHandler` function seems to complicate this term by adding an additional meaning. Can we brainstorm names for this new function?

How about (and I really need some afternoon coffee, to preface this list)

- createHandler
- middlewareChain
- bindMiddleware
- compose ... composeX ... comeposeSomething
- flow (honorable mention)
- E.flow (nope, `E` too often means `Either`)
- stitch

hmm.. I'm trying to avoid touching "handler" at all, which is proving difficult.

What about these name-pieces, if not full names

- event
- responder

hmm. Tough to beat `routeHandler`. Do you have any alternate workarounds to avoid overloading an existing term?

packages/express-wrapper/src/middleware.ts
```diff
@@ -0,0 +1,189 @@
...
+      }
+    });
+
+    result = { ...result, ...middlewareResult };
+  }
+
+  return result as MiddlewareChainOutput<Input, Chain>;
+}
+
+export async function runChainIgnoringMiddleware<Input, Chain extends MiddlewareChain>(
```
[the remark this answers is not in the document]
bitgopatmcl: yeah I can add one, it's mainly to preserve backwards compatibility in the even someone uses one of the fancy new property-appending middlewares as a regular-old express RequestHandler. The eventual handler/serviceFn isn't expecting any properties to be added by middleware, so this is how that happens. This normally wouldn't be an issue except in the case where a middleware clobbers another value (why would you do that? idk).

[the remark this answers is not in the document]
bitgopatmcl: [This test case](https://github.com/BitGo/api-ts/pull/177/files#diff-eead2ab6bba4873d32a5dd6a625479b28d294be1d58dbf6447417f555bd42f63R207-R229) covers it.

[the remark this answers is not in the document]
bitgopatmcl: Renamed and added a docstring

ericcrosson-bitgo (approved):

ericcrosson-bitgo (approved):

PR #179, opened 2022-07-26
fix: add default param to requestHandler impl

bitgopatmcl:
I also figured out how to actually use a narrow impl type without it
complaining.


ericcrosson-bitgo (approved):
Sweet! eslint breathes a sigh of relief :smiley:

PR #184, opened 2022-08-03
chore: Use alejandra for Nix formatting

ghost:
Ticket: BG-00000

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .direnv/
+/.pre-commit-config.yaml
 .drone.yml
 .nyc_output/
 .turbo/
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,6 +1,0 @@
-{
-  "*.json": ["prettier --write"],
-  "*.md": ["prettier --write"],
-  "*.ts": ["prettier --write"],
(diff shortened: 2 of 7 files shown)


ericcrosson-bitgo:
We are already using a linter actually, lint-staged https://github.com/BitGo/api-ts/blob/d4e0640e80abb8be195df70c42ae30b882d31d8b/package.json#L33

I need to check this PR out locally to see that there's no interference with the pre-commit script

ericcrosson-bitgo:
This was incompatible with lint-staged, which is the reason we hadn't migrated to use pre-commit.nix before. I pushed a commit to #185 that we can merge into this PR that removes lint-staged and runs prettier with pre-commit.nix

This may cause difficulties when we add eslint (shortly, I have it planned for this sprint) but that's an egg for a different lunch

ericcrosson-bitgo:
@louib-bitgo Note that this repository uses the fork model; directly pushing branches is not the normal workflow. Maintainers have this ability technically, because we require push permissions to merge PRs, but other contributors will not be able to.

louib-bitgo:
@ericcrosson-bitgo closing this PR in favor of a PR from my own fork.

PR #186, opened 2022-08-04
fix: attempt to handle response types more cleanly

bitgopatmcl:
Redefines a "service function" so that the return type is either all
numeric or all keyed. This is a bit of a guess, but the hope is that it
resolves some of the weird typechecking issues we are seeing around
response types.


jonball4 (approved):

jonball4:
Works well using solely `ResponseType`, can't speak for users of the Keyed counter-part.

ericcrosson-bitgo (approved):

PR #187, opened 2022-08-04
chore: Use alejandra for Nix formatting

ghost:
Also replaces lint-staged with pre-commit.nix

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,9 +10,14 @@
     };
   };
 
-  outputs = { self, nixpkgs, flake-utils, flake-compat }:
-    flake-utils.lib.eachDefaultSystem
-      (system:
+  outputs = {
+    self,
+    nixpkgs,
+    flake-utils,
+    flake-compat,
+  }: (
+    flake-utils.lib.eachDefaultSystem (
+      system: (
         let
           pkgs = nixpkgs.legacyPackages.${system};
         in {
@@ -28,5 +33,7 @@
             '';
           };
         }
-      );
+      )
+    )
+  );
 }
--- a/shell.nix
+++ b/shell.nix
@@ -1,10 +1,14 @@
-(import
+(
+  import
   (
-    let lock = builtins.fromJSON (builtins.readFile ./flake.lock); in
-    fetchTarball {
-      url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
-      sha256 = lock.nodes.flake-compat.locked.narHash;
-    }
+    let
+      lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+    in
+      fetchTarball {
+        url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+        sha256 = lock.nodes.flake-compat.locked.narHash;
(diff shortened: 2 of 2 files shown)


louib-bitgo:
@ericcrosson-bitgo I've just realized that this will remove the pre-commit setup for devs not using Nix. Are we ok with that?

ericcrosson-bitgo:
> @ericcrosson-bitgo I've just realized that this will remove the pre-commit setup for devs not using Nix. Are we ok with that?

That's a great point... I think it's fair to assume this would be a net negative for the average potential contributor. Not that we're flush with those yet, but a team can dream!

Do you think you can wire Nix in to a pre-commit hook using lint-staged?

louib-bitgo:
@ericcrosson-bitgo 

> Do you think you can wire Nix in to a pre-commit hook using lint-staged?

You mean calling a Nix pre-commit hook from lint-stage? I guess I could, but that would still require Nix to be installed/activated, no? And then `eslint` would be called twice.

> That's a great point... I think it's fair to assume this would be a net negative for the average potential contributor. Not that we're flush with those yet, but a team can dream!

We could document that commit hooks are only available when using Nix

ericcrosson-bitgo:
> You mean calling a Nix pre-commit hook from lint-stage? I guess I could, but that would still require Nix to be installed/activated, no? And then `eslint` would be called twice.

@louib-bitgo err no that was my unclear phrasing, I mean invoking alejandra from lint-staged. I guess having alejandra installed would be the real trick... :thinking: 

> We could document that commit hooks are only available when using Nix

This is true, though it still seems a subpar developer experience. I'm more tempted to apply alejandra once manually and call it a day, especially since it's only a formatter and doesn't provide static analysis

flake.nix
```diff
@@ -1,21 +1,39 @@
+# Get started by following https://github.com/BitGo/bitgo-microservices/blob/develop/docs/NIX.md
 {
   description = "api-ts";
 
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs";
     flake-utils.url = "github:numtide/flake-utils";
+    pre-commit-hooks.url = "github:cachix/pre-commit-hooks.nix";
```
bitgopatmcl, line 8:
```suggestion
    pre-commit-hooks = {
      url = "github:cachix/pre-commit-hooks.nix";
      inputs.nixpkgs.follows = "nixpkgs";
      inputs.flake-utils.follows = "flake-utils";
    };
```

Overriding these transitive dependencies should remove the duplicate nixpkgs and flake-utils from the lockfile and improve performance

ericcrosson-bitgo:
Requesting a temperature check on

> This is true, though it still seems a subpar developer experience. I'm more tempted to apply alejandra once manually and call it a day, especially since it's only a formatter and doesn't provide static analysis

Do you agree?

louib-bitgo:
> This is true, though it still seems a subpar developer experience. I'm more tempted to apply alejandra once manually and call it a day, especially since it's only a formatter and doesn't provide static analysis

Yeah that makes sense. I'll do that and revert the other changes

flake.nix
louib-bitgo: I reverted the pre-commit hook changes from this PR

ericcrosson-bitgo (approved):

Issue #135, opened 2022-06-28
Define middleware to run after the request.


jonball4:
We ought to be able to add middleware that runs after our request handler as well.  Currently we can only add middleware that runs before our request via `app.use` or adding middleware via the `RouteHandler`:

https://github.com/BitGo/api-ts/blob/7dc00a4500104ff0b78b9f81bfcb68037dfe5a12/packages/express-wrapper/src/index.ts#L56

Perhaps `beforeMiddleware` and `afterMiddleware` instead of just `middleware` in the `RouteHandler` definition.

bitgopatmcl:
:+1: currently working a route where this is needed as well

Issue #189, opened 2022-08-04
Middleware implementation prevents use of middleware that requires the `next` function


jonball4:
Taking over the `next` function limits the ability to use middleware that requires passing the next function, which is the case with some functions in bitgo's common library.

https://github.com/BitGo/api-ts/blob/039e72ebd477254275d0e0c779549d6ebc376ad5/packages/express-wrapper/src/middleware.ts#L6-L9

PR #192, opened 2022-08-10
feat: add typed-express-router package

bitgopatmcl:
Initial PR for a lighter weight integration of api-ts specifications with Express. The idea is to provide a relatively unopinionated base layer that provides type-checked requests and is flexible enough for other packages to build off of. This should improve familiarity and DX while simultaneously not reinventing the wheel on much of the functionality the existing router provides.

## Goals

- Define routes in Express that are associated with an api-ts `apiSpec`
- Augment the existing Express request with the decoded request object
- Augment the existing Express response with a type-checked `encode` function
- Allow customization of what to do on decode/encode errors, per-route if desired
- Allow action to be performed after an encoded response is sent, per-route if desired
- Allow routes to be defined with path that is different than the one specified in the
  `httpRoute` (e.g. for aliases)
- Follow the express router api as closely as possible otherwise

## Non-Goals

- Enforce that all routes listed in an `apiSpec` have an associated route handler
- Layer anything on top of the `express.RequestHandler[]` chain beyond the additional
  properties described above (projects and other libraries can do this)


bitgopatmcl:
Should resolve #135 both with the after-response hook and by not messing with request handler chains. Also should address #189.

packages/typed-express-router/src/errors.ts
```diff
@@ -0,0 +1,18 @@
...
+) {
+  res.status(400).send('decode error IMPROVE THIS MESSAGE').end();
+}
+
+export function defaultOnEncodeError(
+  _err: unknown,
+  _req: express.Request,
+  res: express.Response,
+) {
+  res.status(500).send('Internal server error IMPROVE THIS MESSAGE').end();
```
ericcrosson-bitgo, line 17: Got a ticket for these action items? :grinning:

ericcrosson-bitgo (dismissed):
The main file `index.ts` file looks pretty simple, true to the idea of this new package. Thanks for the high test coverage! Plenty of examples of use made it easy to review this code

Looks great to me, I won't merge after approval in case you wanted to manually add a git tag that semantic-release will pick up as an input in calculating the next package version. I _think_ if we don't add a tag it'll pick some beta form of 1.0.0 (but on the beta dist tag, [FAQ](https://semantic-release.gitbook.io/semantic-release/support/faq#can-i-set-the-initial-release-version-of-my-package-to-0.0.1)). I have no idea what happens when we drop out of beta in that case, whether we land on v1 or v2

packages/typed-express-router/src/errors.ts
bitgopatmcl: Oops, can copy over the ones from `express-wrapper` for now but still will make one for improving those, particularly the decode error one.

ericcrosson-bitgo (approved):

PR #193, opened 2022-08-11
fix: use methods from io-ts-http

bitgopatmcl:
multi-semantic-release skipped publishing `typed-express-router` because there were no changes between the tag and merge commit. This is a small change that should cause a release.


ericcrosson-bitgo (approved):
Is `Lowercase` a built-in type? Neat-o!

PR #195, opened 2022-08-11
feat: update typed-express-router api

bitgopatmcl:
Quick iteration on the typed-express-router api based on implementation feedback.
Changes:
- Alias methods removed and changed to an option.
- Unchecked method behavior changed, they still attempt to decode a request but just put the resulting `Either` on the request.
- Router-level middleware (`router.use(middleware)`) can inspect the same `Either` decoded requests, even on checked routes before it is validated and unwrapped.

### DIFF
--- a/packages/typed-express-router/README.md
+++ b/packages/typed-express-router/README.md
@@ -5,12 +5,13 @@ A thin wrapper around Express's `Router`
 ## Goals
 
 - Define Express routes that are associated with routes in an api-ts `apiSpec`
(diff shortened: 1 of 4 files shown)


ericcrosson-bitgo (approved):
I like this implementation of aliases. You turned this into a good outcome @bitgopatmcl, breaking through a local maximum with a new way of thinking, and taking it from concept to a far superior solution in a flash. Great work!

PR #198, opened 2022-08-25
fix: check for compiler errors before emitting an openapi spec

bitgopatmcl:
Queries the typescript compiler for any errors before proceeding with
openapi generation. This catches issues such as `node_modules` not being
present, as it'll result in errors like "cannot resolve `io-ts`". It
also should preemptively catch other problems that we haven't
anticipated yet. There is a noticeable performance hit from this, but I
feel it is worth the trade-off for correctness.

### DIFF
--- a/packages/openapi-generator/corpus/test-single-route.ts
+++ b/packages/openapi-generator/corpus/test-single-route.ts
@@ -2,7 +2,6 @@
 
 /// file: other.ts
 
-import * as t from 'io-ts';
 import { NumberFromString } from 'io-ts-types';
 
 /**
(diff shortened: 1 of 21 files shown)


bitgopatmcl:
> There is a noticeable performance hit from this

oof, it is large enough to cause timeouts in CI

ericcrosson-bitgo:
> There is a noticeable performance hit from this

Do we know how this performance hit relates to the size of the input `apiSpec`? 

I ran the tests on my machine, they took 52 seconds to run, which is about 3 times longer than what I remember without this diff. This is slow though tolerable, but our inputs are also fairly trivial. I wonder if we'll leave the "tolerable" zone with a larger input spec

bitgopatmcl:
Test runs down to 1:40-2:00.

ericcrosson-bitgo (approved):
Looks great!

PR #199, opened 2022-08-25
fix: elide empty descriptions when allowed by the spec

bitgopatmcl:
The description field in responses appears to be required. Is there
a reasonable default we can use for these? Perhaps the http status name?

### DIFF
--- a/packages/openapi-generator/corpus/test-array-property.ts
+++ b/packages/openapi-generator/corpus/test-array-property.ts
@@ -33,7 +33,6 @@ export const Routes = h.apiSpec({
     "/test": {
       "get": {
         "summary": "MyRoute",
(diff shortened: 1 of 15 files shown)


ericcrosson-bitgo:
> The description field in responses appears to be required.

Can you tell me more about this? Do you have a doc stating the same?

ericcrosson-bitgo:
Aye, I see it -- thanks for the link!

As discussed in stand-up, I think this is a step towards where we want to be and I agree you've identified another improvement, using a default description by status code (for example: 200 -> "OK", 400 -> "Bad request"). Let's make a ticket to track that work and we'll address it in a future sprint

ericcrosson-bitgo (approved):

PR #202, opened 2022-08-26
feat: constrain api spec keys to HTTP methods

bitgopatmcl:
This also fixes a bit of unsoundness in the types that derived from
`ApiSpec`. Runtime checks are added for things like explicitly undefined
keys on the optional properties (not always preventable depending on
tsconfig).

### DIFF
--- a/packages/express-wrapper/src/index.ts
+++ b/packages/express-wrapper/src/index.ts
@@ -5,7 +5,7 @@
 
 import express from 'express';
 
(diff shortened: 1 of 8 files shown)


bitgopatmcl:
Oops, forgot the additional constraint on the method in the route. Will add real quick

ericcrosson-bitgo (commented):
Looks great! :rocket:

ericcrosson-bitgo (approved):

PR #211, opened 2022-09-01
feat: bump major version of packages out of pre-release

bitgopatmcl:
BREAKING CHANGE: the last attempt failed so trying again.


ericcrosson-bitgo (approved):
Gotta have the scope, good catch

PR #221, opened 2022-09-16
fix: list @types/express as a dependency

ericcrosson-bitgo:
Since the `typed-express-router` exports types from `@types/express` (through [WrappedRouter], defined [here]) we need to make the `@types/express` types visible to consumers of the library. That requires placing `@types/express` in the `dependencies` section of the `typed-express-router` `package.json`, as described by [StackOverflow] and the [TypeScript Documentation].

However, currently `@types/express` is [listed] under `devDependencies`, not `dependencies`.

Closes Ticket: BG-57590

[wrappedrouter]: https://github.com/ericcrosson-bitgo/api-ts/blob/f2fe78bb49ea617d1bea44405601ad694fe88615/packages/typed-express-router/src/index.ts#L29
[here]: https://github.com/ericcrosson-bitgo/api-ts/blob/0898ccecdcebdcf98cff0fded260ae9b8a905825/packages/typed-express-router/src/types.ts#L119
[stackoverflow]: https://stackoverflow.com/a/46011417
[typescript documentation]: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies
[listed]: https://github.com/ericcrosson-bitgo/api-ts/blob/530e1550fe90e25fcbc57429d26f17e54b8fd546/packages/typed-express-router/package.json#L28

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -691,7 +691,6 @@
     },
     "node_modules/@types/body-parser": {
       "version": "1.19.2",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@types/connect": "*",
@@ -706,7 +705,6 @@
     },
     "node_modules/@types/connect": {
       "version": "3.4.35",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@types/node": "*"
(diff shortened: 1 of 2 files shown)


ocornoc (approved):
Makes sense, I appreciate the detailed PR body.

PR #225, opened 2022-09-23
fix(pages): correct the baseUrl configuration

ericcrosson-bitgo:
As suggested by the docusaurus error on the current GitHub Pages
deployment.


ericcrosson-bitgo:
Suggestion visible in this error message 

image

ghost (approved):

PR #228, opened 2022-09-26
ci: use consistent concurrency limits

ericcrosson-bitgo:
For release workflows:

- ensure there is only one workflow running concurrently
- do not skip any workflows

This is because releases act on shared resources (the deployed
environment, or git tags), so this concurrency limit acts as a mutex.

For CI (pre-merge) workflows:

- ensure only one workflow runs per branch
- cancel in-progress workflows

This is because we only care about the results of the latest workflows
to run, which act as required status checks. The impact is faster
reported status-check results, because of less queueing.

Ticket: 0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,10 @@ on:
   push:
     branches: [master, alpha, beta, next, next-major]
 
+# Only allow one release workflow to execute at a time, since each release
+# workflow uses shared resources (git tags, package registries)
+concurrency: ${{ github.workflow }}
+
 jobs:
   release:
     name: Release
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,10 @@ name: CI
 on:
   pull_request:
 
+concurrency:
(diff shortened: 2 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -5,6 +5,10 @@ on:
   push:
     branches: [master, alpha, beta, next, next-major]
 
+# Only allow one release workflow to execute at a time, since each release
+# workflow uses shared resources (git tags, package registries)
+concurrency: ${{ github.workflow }}
```
louib-bitgo, line 10: @ericcrosson-bitgo for consistency, I would use the same syntax as the other actions:
```
concurrency:
  group: ${{ github.workflow }}
```
The shorter syntax is confusing, because the value could be interpreted as a number.

ghost (approved):

.github/workflows/release.yml
ericcrosson-bitgo: That's a great point! I opened https://github.com/BitGo/api-ts/pull/229 to implement this

PR #243, opened 2022-10-04
Upgrade @types/superagent

ericcrosson-bitgo:
This commit adds `@types/node` as a dependency to the root package
manifest.

This pacifies the following compilation error:

```
Error: @api-ts/response:build: ../../node_modules/@types/superagent/index.d.ts(23,10):
          error TS2305: Module '"buffer"' has no exported member 'Blob'.
```

which seems to be caused by a too-old version of `@types/node` being installed.

This too-old version of `@types/node` is installed because 
`@types/superagent` depends on `"@types/node": "*"` (citation
needed, but this is what we see in our `package-lock.json`
file[^1]). This `"*"` means "any installed version will do".
Unfortunately, we have one installed version of `@types/node`:

```
$ npm ls @types/node
└─┬ multi-semantic-release@3.0.1
  └─┬ @manypkg/get-packages@1.1.3
    └─┬ @manypkg/find-root@1.1.0
      └── @types/node@12.20.48
```

which is version 12.20.48, a dependency of multi-semantic-release.

This version of `@types/node` is incompatible with the Node.js runtimes
we're using (currently 14, 16, and 18).

This commit circumvents this old version of node being the one picked
up by `@types/superagent` by forcing a second, more recent version of
`@types/node` to be downloaded and take import-priority over the old
version.

[^1]: https://github.com/BitGo/api-ts/blob/23e2993cf8d5c8e58f9b2bd4b15a2182dacc3fa5/package-lock.json#L798

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -790,9 +790,10 @@
       }
     },
     "node_modules/@types/superagent": {
-      "version": "4.1.5",
+      "version": "4.1.15",
+      "resolved": "https://registry.npmjs.org/@types/superagent/-/superagent-4.1.15.tgz",
+      "integrity": "sha512-mu/N4uvfDN2zVQQ5AYJI/g4qxn2bHB6521t1UuH09ShNWjebTqN0ZFuYK9uYjcgmI0dTQEs+Owi1EO6U0OkOZQ==",
       "dev": true,
-      "license": "MIT",
       "dependencies": {
         "@types/cookiejar": "*",
         "@types/node": "*"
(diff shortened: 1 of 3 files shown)


ericcrosson-bitgo:
Supersedes https://github.com/BitGo/api-ts/pull/242

ericcrosson-bitgo:
Question for the reviewer: can we do without this `@types/superagent` package all together?

bitgopatmcl (approved):

PR #268, opened 2022-11-02
fix: add npm >6 requirement to package.json

bitgopatmcl:
This project uses NPM workspaces that are only supported by version 7 and above

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,5 +39,8 @@
     "prettier": "2.5.1",
     "turbo": "1.6.3",
     "typescript": "4.6.4"
+  },
+  "engines": {
+    "npm": ">6"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -39,5 +39,8 @@
     "prettier": "2.5.1",
     "turbo": "1.6.3",
     "typescript": "4.6.4"
+  },
+  "engines": {
+    "npm": ">6"
```
ericcrosson-bitgo, line 44: Nit: I'm a `>=7` fan myself, for subtle reasons. It seems to bring the focus to the version we want, which makes it easier to recall the why ("what's special about npm 7? Oh right, the workspaces")

ericcrosson-bitgo (approved):

ericcrosson-bitgo:
Thanks @bitgopatmcl !

PR #269, opened 2022-11-02
test(api-ts): remove log in test

jonball4:


ericcrosson-bitgo:
> Ticket: BG-00000

This repository doesn't require tickets, please omit the ticket next time instead of using a fake one.

ericcrosson-bitgo (changes requested):
Do you mind using the `test` conventional-commits type? The `fix` type will cause another release, but we don't have cause to bump another npm version

ericcrosson-bitgo (approved):
Thanks @jonball4 !

Issue #272, opened 2022-11-08
can't use optional with query\params\body props when exactOptionalPropertyTypes = true


adrian-gierakowski:
To reproduce:

make the following changes: https://github.com/rhinofi/api-ts/commit/fc7d71aa4ac34f58abb0a992c0549947cf80d67a

compiler will complain [here](https://github.com/rhinofi/api-ts/blob/fc7d71aa4ac34f58abb0a992c0549947cf80d67a/packages/superagent-wrapper/test/request.test.ts#L22) with following error:

```
/home/adrian/api-ts/packages/superagent-wrapper/test/request.test.ts:
   22:3   error   Type 'Type<{ foo?: string | undefined; id: number; bar: number; }, { query: { foo?: string | undefined; }; params: { id: string; }; body: { bar: number; }; }, unknown>' is not assignable to type 'HttpRequestCodec<any>'. ​typescript:2322
                    Types of property 'encode' are incompatible.
                      Type 'Encode<{ foo?: string | undefined; id: number; bar: number; }, { query: { foo?: string | undefined; }; params: { id: string; }; body: { bar: number; }; }>' is not assignable to type 'Encode<any, { query: { [x: string]: string | string[]; }; params: { [x: string]: string; }; } & { headers?: { [x: string]: string; } | undefined; body?: Json | undefined; }>'.
                        Type '{ query: { foo?: string | undefined; }; params: { id: string; }; body: { bar: number; }; }' is not assignable to type '{ query: { [x: string]: string | string[]; }; params: { [x: string]: string; }; } & { headers?: { [x: string]: string; } | undefined; body?: Json | undefined; }'.
```

for reference:
https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes
https://www.typescriptlang.org/tsconfig#strictNullChecks

adrian-gierakowski:
[this](https://github.com/rhinofi/api-ts/commit/9847c48da18fdfa225033f4c894f658ae422197b) fixes the error, but I'm not sure if it  could have any undesired effects

additional fixes would be required for `params`, `body` and possibly `headers`

bitgopatmcl:
Thanks for the bug report! As far as I can see, your fix would work, but it would be nice if we could keep the `GenericHttpRequest` type as simple as possible. I've opened #274 as a possible alternative solution that also enables `optionalized` to take advantage of the benefits of `exactOptionalPropertyTypes`. Please take a look and let me know if it works for your use case.

[...]

PR #274, opened 2022-11-11
BREAKING CHANGE: make `optionalized` strip explicit `undefined` values

bitgopatmcl:
`optionalized` determines which properties to make optional by checking if the codec associated with the property accepts `undefined`. As a result, it is impossible to use it to define types that have optional properties whose type isn't a union containing `undefined`. When `exactOptionalPropertyTypes` is `false` this isn't a big deal, because all optional properties are implicitly treated this way. However, it would be nice to be able to take advantage of the cleaner types enabled by `exactOptionalPropertyTypes`.

This change offers a potential solution by changing the behavior of `optionalized` such that it always deletes any properties that are explicitly set to `undefined`. This is a breaking change, but seems unlikely to cause issues in practice. If the previous behavior is needed, it is always possible to use
`t.intersection([t.type(...),t.partial(...)])`.

This change also solves an inconsistency in `io-ts-http` exposed when `exactOptionalPropertyTypes` is enabled, as reported in #272. The `httpRequest` function is implemented with `flatten`, which currently passes all keys to `optionalized`. At the same time, `GenericHttpRequest` is defined using `Record` types whose codomains do not include `undefined`. With `exactOptionalPropertyTypes` enabled, this results in an error if, for example, an optional query parameter is defined on an `httpRequest` because of how `optionalized` forces all optional properties to be able to be explicitly `undefined`. Rather than adding `t.undefined` to all the records, this change presents an alternative that keeps the `GenericHttpRequest` type as simple as possible.

### DIFF
--- a/packages/io-ts-http/src/combinators.ts
+++ b/packages/io-ts-http/src/combinators.ts
@@ -3,6 +3,7 @@ import * as E from 'fp-ts/Either';
 import * as R from 'fp-ts/Record';
 import * as t from 'io-ts';
 import {
+  DefinedValues,
   Flattened,
   NestedType,
   NestedOutputType,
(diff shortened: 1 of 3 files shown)


packages/io-ts-http/src/combinators.ts
```diff
@@ -13,10 +14,51 @@ import {
...
+> => {
+  const partialCodec = t.partial(props, name);
+  return new t.PartialType(
+    partialCodec.name,
+    (i): i is DefinedValues<t.TypeOfPartialProps<P>> => partialCodec.is(i),
+    (i, ctx) => {
+      return pipe(
+        partialCodec.validate(i, ctx),
+        E.map((result) => {
+          for (const key in result) {
```
ericcrosson-bitgo, line 33: Should we use `Object.keys` here, which only enumerates owned properties?

```suggestion
          for (const key of Object.keys(result)) {
```

packages/io-ts-http/src/combinators.ts
```diff
@@ -13,10 +14,51 @@ import {
...
+              delete result[key];
+            }
+          }
+          return result as DefinedValues<t.TypeOfPartialProps<P>>;
+        }),
+      );
+    },
+    (a) => {
+      const result = partialCodec.encode(a);
+      for (const key in result) {
```
ericcrosson-bitgo, line 44: idem

ericcrosson-bitgo (commented):
Just did a quick pass, commenting to post initial feedback. I'm going to fetch this code and play with it locally

packages/io-ts-http/src/combinators.ts
ericcrosson-bitgo: Similar to this change: https://github.com/BitGo/api-ts/pull/277

ericcrosson-bitgo:
Seeing the 
```typescript
if (result[key] === undefined) {
```
reminds me that this doesn't always return what you think it will, since it's possible to override `undefined`[^1] :sweat_smile: 

Instead, if we were the paranoid type, we could use the fail-safe 

```typescript
if (result[key] === void 0) {
```

I'm not requesting changes, just thought I'd share!

[^1]: https://javascript.plainenglish.io/the-different-ways-to-check-for-undefined-acaaa385a719

packages/io-ts-http/src/combinators.ts
```diff
@@ -13,10 +14,51 @@ import {
...
+};
+
 export const optional = <C extends t.Mixed>(subCodec: C) =>
   t.union([subCodec, t.undefined]);
 
-export const optionalized = <P extends t.Props>(props: P): OptionalizedC<P> => {
+export const optionalized = <P extends t.Props>(
+  props: P,
+  name?: string,
+): OptionalizedC<P> => {
```
ericcrosson-bitgo, line 61: Accepting `name` here seems worth calling out in the commit message/release notes!

packages/io-ts-http/src/utils.ts
```diff
@@ -9,12 +21,12 @@ export type PossiblyUndefinedProps<T extends t.Props> = {
 }[keyof T];
 
 type Optionalized<T> = Simplify<
-  Omit<T, PossiblyUndefinedKeys<T>> & Partial<Pick<T, PossiblyUndefinedKeys<T>>>
+  Omit<T, PossiblyUndefinedKeys<T>> &
+    Partial<DefinedValues<Pick<T, PossiblyUndefinedKeys<T>>>>
```
ericcrosson-bitgo, line 25: Gnarly! :man_playing_handball:

ericcrosson-bitgo (dismissed):
Looks beautiful, nice work!

bitgopatmcl:
Had to resolve a merge conflict, so took the opportunity to add an explicit `unknown` type parameter and change comparison to `void 0`

ericcrosson-bitgo (approved):

PR #279, opened 2022-11-11
refactor: use OutputOf to mirror expected order of type arguments

ericcrosson-bitgo:
Change the type-level function wrapping the second type-argument to
`t.Type` from `TypeOf` to `OutputOf`, to match the expected order
of io-ts type-arguments: `A, O, I`.

This has no change in type or behavior, since in this case `TypeOf`
and `OutputOf` behave the same. However, this change should improve
readability.


bitgopatmcl (approved):
🚀

PR #283, opened 2022-11-15
chore: run `npm audit fix`

bitgopatmcl:
Not sure why dependabot missed these ones.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4444,9 +4444,10 @@
       }
     },
     "node_modules/minimatch": {
-      "version": "3.0.4",
+      "version": "3.1.2",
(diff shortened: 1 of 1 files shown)


ericcrosson-bitgo:
Seems like dependabot supports transitive npm dependencies? https://github.blog/changelog/2022-09-07-dependabot-unlocks-transitive-dependencies-for-npm-projects/

ericcrosson-bitgo:
Ah, we have to turn on Dependabot security updates. @bitgopatmcl will you check if there's a way to do that from terraform?

ericcrosson-bitgo (approved):

PR #308, opened 2022-11-21
chore: remove shell.nix file

ericcrosson-bitgo:
After talking to @louib-bitgo, we don't have a use case for this file
and so it can be deleted.


ghost (approved):
@ericcrosson-bitgo I'm pretty sure we can also remove the `flake-compat` import from the flake. See https://github.com/edolstra/flake-compat#flake-compat

ericcrosson-bitgo:
@louib-bitgo Thanks for the tip, implemented in https://github.com/BitGo/api-ts/pull/309

PR #355, opened 2022-12-14
fix: expose middleware handler

jonball4:
Access to the middleware handler type is a necessity for building type safe middleware chains

### DIFF
--- a/packages/express-wrapper/src/middleware.ts
+++ b/packages/express-wrapper/src/middleware.ts
@@ -8,7 +8,7 @@ export type MiddlewareFn<T extends {}> = (
   res: express.Response,
 ) => Promise<T>;
 
-type MiddlewareHandler<T extends {} = {}> = {
+export type MiddlewareHandler<T extends {} = {}> = {
   (req: express.Request, res: express.Response, next: express.NextFunction): void;
   [MiddlewareBrand]: T;
 };


ericcrosson-bitgo:
@jonball4 can you provide more detail? What are you trying to do that isn't working without this identifier exported?

I'm asking because each export becomes a long-term commitment, and we'd like to make sure the value added exceeds the cost of that commitment.

ericcrosson-bitgo (commented):
@jonball4 can you provide more detail? What are you trying to do that isn't working without this identifier exported?

I'm asking because each export becomes a long-term commitment, and we'd like to make sure the value added exceeds the cost of that commitment.

jonball4:
> @jonball4 can you provide more detail? What are you trying to do that isn't working without this identifier exported?
> 
> I'm asking because each export becomes a long-term commitment, and we'd like to make sure the value added exceeds the cost of that commitment.

@ericcrosson-bitgo understood.  Here is a link to the code I'm trying to pull into bitgo-network:

https://github.com/BitGo/logging/blob/beta/src/middleware/express-wrapper.ts

I would like to be able to annotate the return types here and I can't do that properly without access to the internal MiddlewareHandler type.  There is a MiddlewareRequestHandler type but that is a union between an express request and a MiddlewareHandler with no generic arguments, leading to a middleware chain output type that has an empty object as its return type rather than the data I'm trying to pass to my handlers.

ericcrosson-bitgo:
Thanks for the context. It seems reasonable to annotate a function's return types if we extract it from a middleware chain

ericcrosson-bitgo (approved):

PR #356, opened 2022-12-15
fix: relax io-ts-types constraint

jonball4:
The existing constraint on io-ts-types version 0.5.19 prevents upgrading without resolving loads of package dependency issues in monorepos using 0.5.16 for example.

### DIFF
--- a/packages/io-ts-http/package.json
+++ b/packages/io-ts-http/package.json
@@ -22,7 +22,7 @@
     "@api-ts/response": "0.0.0-semantically-released",
     "fp-ts": "^2.0.0",
     "io-ts": "2.1.3",
-    "io-ts-types": "0.5.19"
+    "io-ts-types": "^0.5.16"
   },
   "devDependencies": {
     "@types/chai": "4.3.4",
--- a/packages/openapi-generator/package.json
+++ b/packages/openapi-generator/package.json
@@ -25,7 +25,7 @@
     "comment-parser": "1.3.1",
     "fp-ts": "2.13.1",
     "io-ts": "2.1.3",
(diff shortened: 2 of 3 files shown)


ericcrosson-bitgo:
@jonball4 how did you choose 0.5.16?

ericcrosson-bitgo (changes requested):
@jonball4 how did you choose 0.5.16?

I suspect there should be lockfile changes too

jonball4:
> @jonball4 how did you choose 0.5.16?
> 
> I suspect there should be lockfile changes too

oo thanks, need to run an install.  0.5.16 is what is used widely in the monorepo.

ericcrosson-bitgo (commented):
I propose we supersede this PR with #358 for the following reasons:

- 0.5.15 is the minimum-supported io-ts-types version

  The monorepo's current state should not factor in to decisions made in this repository

- superagent-wrapper does not need to be updated, since it imports io-ts-types as a devDependency

  devDependencies are not downloaded by consumers of this library, they are only downloaded when developing this library directly. There will be no conflict with io-ts-types imported by superagent in other packages, and relaxing the semver constraint in this case also weakens our guarantee of "we can run tests with superagent-wrapper with this commit" to "the last time I tried running tests on this commit, it worked"

- openapi-generator is a binary, not a library, and we prefer exact versions over semver ranges for strength of reproducibility guarantees as mentioned in the previous bullet

  It's possible the current installation procedure for openapi-generator will cause similar issues down the road, but we might want to address that in a different way. If this package isn't causing an immediate issue, I suggest we leave it alone for now

PR #362, opened 2022-12-16
chore(dependabot): disable transitive upgrades

ericcrosson-bitgo:
The level of churn introduced by indirect dependency upgrades is
introducing an unwelcome amount of toil and inflating the repository
size.


ericcrosson-bitgo:
Superseded by #363

PR #378, opened 2023-01-17
chore(deps): remove unused dependency

ericcrosson-bitgo:
@semantic-release/npm is included in semantic-release
(which itself is a transitive dependency of multi-semantic-release).
Additionally, we aren't using @semantic-release/npm, we're using
@semrel-extra/npm which wraps @semantic-release/npm with a slight
patch which prevents npm from failing release workflows with an
error due to rate-limiting in a monorepo-release configuration.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,6 @@
         "packages/**"
       ],
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


bitgopatmcl (approved):
Removes `@semantic-release/npm`

PR #380, opened 2023-01-18
fix(superagent-wrapper): support non-`/` api base paths

### LINKED ISSUE #337: Base path is lost when replacing path params
When path params get replaces, we loose any base path that was set in the base url.

https://github.com/BitGo/api-ts/blob/master/packages/superagent-wrapper/src/request.ts#L79-L90

For example, if the base is `http://localhost/foo` and the route path is `/bar`
The url pathname contains `/foo`  at the beginning and after replacing the params in `/foo` the `/bar` part is lost.

That means that if all our endpoints have a common base param we can pass it in the base URL and we need to add it to every route.

Could we just concatenate the result of `substitutePathParams` to `url.pathname` instead of replacing it?

bitgopatmcl:
This should fix #337. I initially tried to use an alternate URL constructor like:
```
const url = new URL(substitutedPath, basePath);
```
and was surprised to learn that this actually has the exact same behavior of stripping the base path. It seems like it should be possible to make the substituted path a relative one, but that [has issues as well](https://github.com/whatwg/url/issues/531).

Ultimately I pulled in `posix.join` from `path`. This works, but may potentially introduce issues with some web bundlers if they don't provide an implementation of Node's `path` module.

### DIFF
--- a/packages/superagent-wrapper/src/request.ts
+++ b/packages/superagent-wrapper/src/request.ts
@@ -1,9 +1,10 @@
 import * as h from '@api-ts/io-ts-http';
 import * as E from 'fp-ts/Either';
+import { pipe } from 'fp-ts/pipeable';
 import * as t from 'io-ts';
 import * as PathReporter from 'io-ts/lib/PathReporter';
+import { posix } from 'path';
 import { URL } from 'whatwg-url';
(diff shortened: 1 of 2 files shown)


ericcrosson-bitgo (approved):
Preserves the base url after path-param substitution :rocket:

PR #387, opened 2023-01-24
chore(superagent-wrapper): move fp-ts to a peer dependency

anshuldoshi:
Packages using superagent-wrapper also typically use fp-ts. Moving fp-ts to a peer dependency of superagent-wrapper to avoid collisions of fp-ts versions between superagent-wrapper and packages that use superagent-wrapper

Ticket: BG-66089

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10476,7 +10476,6 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@api-ts/io-ts-http": "0.0.0-semantically-released",
-        "fp-ts": "^2.0.0",
         "io-ts": "2.1.3",
         "whatwg-url": "12.0.0"
       },
@@ -10497,6 +10496,9 @@
         "supertest": "6.3.3",
         "ts-node": "10.9.1",
         "typescript": "4.7.4"
+      },
+      "peerDependencies": {
+        "fp-ts": "^2.0.0"
       }
     },
     "packages/superagent-wrapper/node_modules/tr46": {
@@ -10692,7 +10694,6 @@
         "c8": "7.12.0",
         "chai": "4.3.7",
         "express": "4.18.2",
-        "fp-ts": "^2.0.0",
         "io-ts": "2.1.3",
         "io-ts-types": "0.5.19",
         "mocha": "10.2.0",
--- a/packages/superagent-wrapper/package.json
+++ b/packages/superagent-wrapper/package.json
@@ -18,7 +18,6 @@
   },
   "dependencies": {
     "@api-ts/io-ts-http": "0.0.0-semantically-released",
-    "fp-ts": "^2.0.0",
     "io-ts": "2.1.3",
     "whatwg-url": "12.0.0"
   },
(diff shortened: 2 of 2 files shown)


ericcrosson-bitgo (changes requested):
@anshuldoshi I don't believe this change is desirable, because superagent-wrapper imports from fp-ts directly[^1]. This diff may cause regressions for Node.js 14 users, since npm 6 does not automatically resolve peer dependencies.

That's a fancy way of saying, npm 6 does not download peer dependencies that happen to not be specified as dependencies too. 

But let's examine that case -- if we're a service importing the superagent-wrapper, and we depend on fp-ts directly, we're combining our own fp-ts semver range with superagent-wrapper's semver range (`^2.0.0`). As long as the service is using fp-ts v2 (the latest major version of fp-ts), there will only be one copy of fp-ts downloaded because the `^2.0.0` range includes every v2 release (so it can be deduplicated). 

So I can't see one case of inputs that would actually cause a duplicate fp-ts version to be downloaded, which means moving to peer dependencies won't actually change behavior.

Do you have a reproducible scenario where this diff produces a change in behavior?

[^1]: https://github.com/BitGo/api-ts/blob/master/packages/superagent-wrapper/src/request.ts#L2-L3

anshuldoshi:
@ericcrosson-bitgo thanks for your feedback!

I agree this would cause issues for users who use npm <7. I thought this shouldn't be an issue though because in the project's base package.json npm >=7 is specified https://github.com/BitGo/api-ts/blob/master/package.json#L42-L43. Not sure how this is handled when other projects that use npm <7 import superagent-wrapper as a dependency though. Is it viable to add fp-ts as both a peerDependency and a dependency?

Your point about the fp-ts versions effectively being the same in superagent-wrapper and packages that use superagent-wrapper also makes sense to me. However, making a [similar change to schema-decoding ](https://github.com/BitGo/schema-decoding/pull/56/files) has earlier this week solved a similar issue to what I am seeing now.

For context, I am running into [build issues](https://drone.bitgo-dev.com/BitGo/bitgo-microservices/207093/3/54) for [this commit](https://github.com/BitGo/bitgo-microservices/pull/28080/commits/4c1d62bfec800b4dfba0477a436722557bde6d97#diff-12ebac6b535ddf60f620ea09facc5846353b8ca2a42dcbe57c072b5875857b13). The problems appear to be coming from superagent-wrapper because when removing usage of this package from the change, build issues disappear. I saw that like schema-decoding, superagent-wrapper was using its own version of fp-ts and believed this to be the issue. Schema-decoding was also using a 2.x.x version of fp-ts yet (for reasons I don't fully understand) this did resolve the issue. Perhaps I am misinterpreting the schema-decoding bug fix but hopefully that explains the motivation behind the change.

I have not tested out if this change to superagent-wrapper does actually resolve my issue, mainly because I am not sure how to do so without an incremented version of superagent-wrapper to test with.

anshuldoshi:
closing because another solution was found

PR #394, opened 2023-02-07
fix: node16 request decoding for typed-express-router

jrball3:
With the same version of express, and io-ts, running on node 16 and decoding a request with a simple `headers: { authorization: string }` specification for an `httpRoute`, an error complaining about `headers` being `undefined` is returned alongside a 400 error for properly constructed requests.

I can't fully explain why this PR fixes the problem, frankly it doesn't make any sense to me at all, but decoding the full request object consistently yields an `E.Left`, while decoding only the relevant properties results in an `E.Right`.  Given the definition of an `HttpRoute`, this change should have no effect other than fixing this strange bug.

```
export const v1BusinessListRoute = httpRoute({
  path: v1BusinessesListPath,
  method: 'GET',
  request: httpRequest({
    query: {
      businessIds: optional(arrayFromQueryParam(UUID)),
      names: optional(arrayFromQueryParam(t.string)),
    },
    params: {},
    headers: { authorization: t.string },
    body: {}
  }),
  response: {
    [HttpStatusCodes.OK]: V1BusinessListResponse,
    [HttpStatusCodes.BAD_REQUEST]: V1ErrorResponseC,
    [HttpStatusCodes.NOT_FOUND]: V1ErrorResponseC,
    [HttpStatusCodes.INTERNAL_SERVER_ERROR]: V1ErrorResponseC,
    [HttpStatusCodes.UNAUTHORIZED]: V1ErrorResponseC,
  }
});
```

```
  const response = await v1BusinessesListClient.get({
    authorization: t.context.authorization,
    businessIds: [t.context.business1.id, t.context.business2.id],
    names: [t.context.business1.name],
  }).decodeExpecting(HttpStatusCodes.OK);
```

```
    message: 'Unexpected response 400: {"error":{"code":"E005","message":"Invalid value undefined supplied to : httpRequest/headers: Exact<({ authorization: string } & Partial<{  }>)>","request":{"headers":{"host":"127.0.0.1:62194","accept-encoding":"gzip, deflate","authorization":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VySWQiOiJlMzY1MTRlZi1hNTFjLTRjNjYtYWJlYi04NTMwZjQ1NWRlOTYiLCJpYXQiOjE2NzU3NDI3ODksImV4cCI6MTY3NTgyOTE4OX0.8wfkn7HoCsn88I2JcUI4hS6ZSd9nAA9T3vwewMRKubs","content-type":"application/json","content-length":"2","connection":"close"},"params":{},"query":{"businessIds":"ea8fc691-540e-4012-8a3a-05618d516d68"},"path":"/api/v1/businesses","body":{}}}}',
```

### DIFF
--- a/packages/typed-express-router/src/index.ts
+++ b/packages/typed-express-router/src/index.ts
@@ -1,11 +1,11 @@
 /*
  * @api-ts/typed-express-router
  */
-
 import { ApiSpec, HttpRoute, KeyToHttpStatus } from '@api-ts/io-ts-http';
 import express from 'express';
 import * as E from 'fp-ts/Either';
 import { pipe } from 'fp-ts/pipeable';
+
 import { defaultOnDecodeError, defaultOnEncodeError } from './errors';
 import { apiTsPathToExpress } from './path';
 import {
@@ -87,7 +87,12 @@ export function wrapRouter<Spec extends ApiSpec>(
         throw Error(`Method "${method}" at "${apiName}" must not be "undefined"'`);
       }
       const wrapReqAndRes: UncheckedRequestHandler = (req, res, next) => {
-        const decoded = route.request.decode(req);
+        const decoded = route.request.decode({
+          body: req.body,
+          headers: req.headers,
+          params: req.params,
+          query: req.query,
+        });
         req.decoded = decoded;
         req.apiName = apiName;
         req.httpRoute = route;


jonball4:
@BitGo/velocity this is my personal github account if not already obvious

ericcrosson-bitgo:
> @BitGo/velocity this is my personal github account if not already obvious

Thank you for the proof of life :grin: We'll need to gpg sign every commit before merge -- do you mind re-opening this PR from @jonball4?

Additionally, since we're all so confused by this behavior, I wonder (slightly) if it's possible to create a test to ensure we don't regress... hmm... probably not worth it, but a comment linking back to this PR with its great description would be much appreciated!

jonball4:
> > @BitGo/velocity this is my personal github account if not already obvious
> 
> Thank you for the proof of life 😁 We'll need to gpg sign every commit before merge -- do you mind re-opening this PR from @jonball4?
> 
> Additionally, since we're all so confused by this behavior, I wonder (slightly) if it's possible to create a test to ensure we don't regress... hmm... probably not worth it, but a comment linking back to this PR with its great description would be much appreciated!

https://github.com/BitGo/api-ts/pull/394 :)

PR #400, opened 2023-02-11
feat(response)!: keyed/non-keyed responses 

jonball4:
Changes
-
- Rename the "Response" functions to "KeyedResponse" functions to align with conventions within express-wrapper. These functions use string-based "type" values alongside the payload.
- Add "Response" functions in place of the original, further aligning with conventions within express-wrapper.  These functions use numerical-based "type" values alongside the payload.
- Add "conflict" / 409 error code to the possible status that can be returned

BREAKING CHANGE: Existing response functions have been renamed to align convention with other packages in this repository.


packages/response/src/response.ts
```diff
@@ -0,0 +1,35 @@
...
+// 401 (unauthorized)          | UNAUTHENTICATED    | Response.unauthenticated
+// 403 (forbidden)             | PERMISSION_DENIED  | Response.permissionDenied
+// 404 (not found)             | NOT_FOUND          | Response.notFound
+// 405 (method not allowed)    | NOT_FOUND          | Response.notFound
+// 409 (conflict)              | ALREADY_EXISTS     | Response.conflict
+// 429 (rate-limit)            | RESOURCE_EXHAUSTED | Response.rateLimitExceeded
+// 500 (internal server error) | INTERNAL           | Response.internalError
+// 503 (service unavailable)   | UNAVAILABLE        | Response.serviceUnavailable
+
+export type Status = 200 | 400 | 401 | 403 | 404 | 409 | 429 | 500 | 503;
```
manav-c, line 17: not a change req but wanted to point out that we specify 405 in the comment above but can't actually throw it

packages/io-ts-http/src/statusCode.ts
```diff
@@ -6,6 +6,7 @@ export const HttpToKeyStatus = {
   401: 'unauthenticated',
   403: 'permissionDenied',
   404: 'notFound',
+  409: 'conflict',
```
manav-c, line 9: 🎸

manav-c (dismissed):
LGTM

packages/response/src/response.ts
jonball4: good catch, this was here before fwiw

manav-c: yeah I figured, just wanted to point it out

bitgopatmcl (approved):
Looks like this breaking change should help with some of the confusing errors that can be generated around responses.

PR #404, opened 2023-02-15
feat(openapi-generator): ability to output yaml specs if needed

manav-c:
Ticket: BG-66640. 

Changes: 
- Added the ability to output a yaml version of the OpenAPI spec if the output file (specified using the --output option on the cli) has a .yaml extension.


ericcrosson-bitgo (changes requested):
Thank you for the suggestion @manav-c, but we have intentionally omitted yaml support to keep our use of third-party dependencies minimal. Let's confer in a Jira ticket if this is more than just a convenience.

PR #412, opened 2023-02-22
chore(deps): use the express-bundled body-parser

ericcrosson-bitgo:
Remove an explicit dependency on body-parser since express includes
its own version.

Supersedes #411


bitgopatmcl (approved):
Removes redundant dependency

PR #425, opened 2023-03-17
chore(dependabot): disable typescript 5 upgrades

ericcrosson-bitgo:
io-ts 2.1.3 is not compatible with typescript >4.8, so improve the
semver range disabling automated upgrade PRs to include typescript 5 and
higher (for now).

This is not where we want to keep api-ts long term, so regard this as
a medium-term solution.

Supersedes #424


bitgopatmcl (approved):
Ignores TS `>4.8`

PR #432, opened 2023-03-24
api-ts openapi-generator updates

manav-c:
[feat(generator): parse route operationId, platform tags, summary](https://github.com/BitGo/api-ts/commit/e06fe9476a123492306825bc97faac65783fdc34)
- Ability to parse `@operationId` jsDoc tag on `httpRoute` definitions. This populates the URL in the `dev-portal`.
- Ability to parse `@summary` jsDoc tag on `httpRoute` definitions. Previously the summary in the openAPI spec would be the name given to the `httpRoute`. This update lets you set a summary. The summary populates the LHS navigation for the route in the `dev-portal`.
- Ability to parse `@tags` jsDoc tag on `httpRoute` definitions. These are the tags that go into `platform.yaml` in the `api-docs` package and under your kit name in `dev-portal`. This allows for proper grouping/classification of your routes on the developer portal.

[feat: parse nonEmptyArrayFromQueryParam, schema for optional](https://github.com/BitGo/api-ts/commit/d95a4d738b0f5ca365b51be17e94066a059b458a)
- Ability to parse `nonEmptyArrayFromQueryParam` codec in request bodies. 
- Fixed the schema for the `optional` codec. Previously `optional` would translate to an array type which is wrong.

### DIFF
--- a/packages/openapi-generator/src/expression.ts
+++ b/packages/openapi-generator/src/expression.ts
@@ -174,6 +174,21 @@ const SPECIAL_CASES: SpecialCaseMap = {
         })),
       ),
   },
+  'query-param-types/dist/src/index': {
+    nonEmptyArrayFromQueryParam: (location) =>
+      pipe(
+        getFirstArgument(location),
+        RE.chain(toOpenAPISchema),
+        RE.map(({ schema }) => ({
+          schema: {
+            title: 'NonEmptyArrayFromQueryParam',
+            type: 'array',
+            items: schema,
+          },
+          required: true,
+        })),
+      ),
+  },
   'io-ts-types/lib/DateFromISOString': {
     DateFromISOString: () =>
       RE.right({
@@ -190,7 +205,7 @@ const SPECIAL_CASES: SpecialCaseMap = {
         getFirstArgument(location),
         RE.chain(toOpenAPISchema),
         RE.map(({ schema }) => ({
-          schema: { type: 'array', items: schema },
+          schema,
           required: false,
         })),
       ),
--- a/packages/openapi-generator/src/project.ts
+++ b/packages/openapi-generator/src/project.ts
@@ -89,6 +89,8 @@ const routesForSymbol =
               body,
               responses,
               summary,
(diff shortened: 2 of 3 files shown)


manav-c (commented):
didn't realize I had to update tests. Doing that. I've used the tool locally with these updates to get BitGo Network docs published successfully.

packages/openapi-generator/src/expression.ts
```diff
@@ -174,6 +174,21 @@ const SPECIAL_CASES: SpecialCaseMap = {
...
+        RE.map(({ schema }) => ({
+          schema: {
+            title: 'NonEmptyArrayFromQueryParam',
+            type: 'array',
+            items: schema,
+          },
+          required: true,
+        })),
+      ),
+  },
```
ericcrosson-bitgo, line 191: I don't love that this is not a conventional or public type, it's a special case that at best might be confusing to a reader not familiar with this codec or at worst, happen to match against an unintended codec (slim though the odds are! :sweat_smile:)

I don't have a suggestion to improve right now, so just creating a place for conversation and not requesting changes

ericcrosson-bitgo (commented):
Thank you for working on these changes! I like the direction.

Will you please add tests to cover the new functionality you added? The tests are defined declaratively but we had to use a little jank to get them running with the expected parallelization, so let us know if you hit any trouble or have any questions.

packages/openapi-generator/src/expression.ts
manav-c: That's true. The motivation behind adding this is that a lot of our routes (BitGo Network) and I think other routes within BitGo that will use `api-ts` and `io-ts` to their full extent will need to use this codec at some point. We use it for all query params in all our routes.

ericcrosson-bitgo:
Closing for house-keeping

Please note that much of this functionality has been included in [@api-ts/openapi-generator@v2.0.0](https://github.com/BitGo/api-ts/releases/tag/%40api-ts%2Fopenapi-generator%402.0.0)

PR #434, opened 2023-03-29
feat: move Response to io-ts-http

andrew-scott-fischer:
Ticket: BG-71730

Content from the ticket:

The Response library in api-ts is redundant with io-ts-http. Response's contents should be moved under the http specific io-ts-http.

Ask: Move relevant types from Response to io-ts-http and delete Response.


bitgopatmcl (approved):
Implements first stage of follow-up plan [described in this PR](https://github.com/BitGo/api-ts/pull/418#discussion_r1128052529)

PR #437, opened 2023-03-31
chore(dependabot): fix typo in semver range

ericcrosson-bitgo:
This range was intended to ignore all versions greater than 4.8, but
there is a typo that renders the range invalid.


ericcrosson-bitgo:
Supersedes https://github.com/BitGo/api-ts/pull/436

ghost (approved):

PR #467, opened 2023-05-28
test: replace ava with node:test

ericcrosson-bitgo:
This PR replaces ava with the Node.js native test runner[^1].

The refactor was straightforward in the express-wrapper and the typed-express-wrapper, but there are some notable changes in the openapi-generator.

The naive implementation ran each test in the test corpus serially, due to this nature of the native test runner's execution model[^2]:

> Each test file is executed as if it was a regular script. That is,
> if the test file itself uses node:test to define tests, all of those
> tests will be executed within a single application thread, regardless
> of the value of the concurrency option of test().

Since the function under test is synchronous and compute-bound, and Node.js is single-threaded, all the tests were executing serially and took over 30s to complete on my macbook.

Consequently, this commit uses Node.js worker threads[^3] to spawn one worker per test-corpus file, limited to one worker per CPU core. In this configuration, all tests complete in 17s on my macbook.

Supersedes #465

[^1]: https://nodejs.org/api/test.html
[^2]: https://nodejs.org/api/test.html#test-runner-execution-model
[^3]: https://nodejs.org/api/worker_threads.html


.github/workflows/ci.yml
ericcrosson-bitgo, line 1: Please note this diff, where we necessarily stop running unit tests on Node.js 14 and 16. I personally am all right with this change, but would love to hear what you think!

bitgopatmcl (approved):
This works :shipit:

PR #499, opened 2023-08-02
test: replace mocha with node:test

ericcrosson-bitgo:
This PR replaces mocha with the Node.js native test runner[^1].

[^1]: https://nodejs.org/api/test.html

### DIFF
--- a/packages/superagent-wrapper/test/request.test.ts
+++ b/packages/superagent-wrapper/test/request.test.ts
@@ -1,5 +1,8 @@
+import { strict as assert } from 'node:assert';
+import { before, after, describe, it } from 'node:test';
+import * as http from 'node:http';
+
 import * as h from '@api-ts/io-ts-http';
-import { assert } from 'chai';
 import express from 'express';
 import * as E from 'fp-ts/Either';
 import * as t from 'io-ts';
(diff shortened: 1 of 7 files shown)


packages/superagent-wrapper/test/request.test.ts
```diff
@@ -62,236 +65,264 @@ const TestRoutes = h.apiSpec({
...
-    });
-  } else {
-    const response = PostTestRoute.response[200].encode({
-      ...params,
-      baz: true,
-    });
-    res.send(response);
-  }
-});
+const createTestServer = (port: number) => {
```
bitgopatmcl, line 68: If the port is hardcoded per-test now, do we even need supertest? Or can we just use express and superagent directly?

bitgopatmcl (approved):
Tested and it works 🎉 . Is the main change in the `superagent-wrapper` tests to instantiate a server in the before block of each section with a hardcoded port, or did I miss something else?

packages/superagent-wrapper/test/request.test.ts
bitgopatmcl: Ah right, supertest also provides some extra functionality on the request object

ericcrosson-bitgo:
> Is the main change in the `superagent-wrapper` tests to instantiate a server in the before block of each section with a hardcoded port, or did I miss something else?

That's the gist of it!

PR #520, opened 2023-08-24
chore(dependabot): group @swc/core-* updates into one PR

ericcrosson-bitgo:
We aim for @swc/core and associated manually-installed optional dependencies to remain pinned to the same version, but the current dependabot configuration opens two separate PRs: #513 #515
 
 This commit configures dependabot to group the @swc/core* updates into a single PR, to keep all @swc/core dependencies on the same version at all times.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,6 +23,10 @@ updates:
     allow:
       - dependency-type: direct
     versioning-strategy: increase-if-necessary
+    groups:
+      dependencies:
+        patterns:
+          - '@swc/core-*'
     ignore:
       # Ignore all @types/nodes patch updates, since
       # this package updates so frequently


.github/dependabot.yml
```diff
@@ -23,6 +23,10 @@ updates:
     allow:
       - dependency-type: direct
     versioning-strategy: increase-if-necessary
+    groups:
+      dependencies:
+        patterns:
+          - '@swc/core-*'
```
bitgopatmcl, line 29: Since the base `@swc/core` library and the native `@swc/core-*` optional dependencies appear to be released in sync, should this be `@swc/core*` instead?

ericcrosson-bitgo: I believe so! Good catch, thank you

bitgopatmcl (approved):

PR #525, opened 2023-08-25
ci(deps-dev): group swc core and native dependabot updates

bitgopatmcl:
Makes the change described in https://github.com/BitGo/api-ts/pull/520#discussion_r1304579036


ericcrosson-bitgo (approved):
I must have fat-fingered the amend -- thanks for fixing this!

PR #536, opened 2023-09-05
feat: output OpenAPI v3 instead of v3.1

bitgopatmcl:
It appears that a lot of tooling does not support OpenAPI v3.1, so this changes `openapi-generator` to output OpenAPI v3 for broader compatibility. The largest change is with explicit `null`-typed properties, since OpenAPI does not have a `null` primitive type. As a workaround, I have it emit a `nullable` type with an empty `enum`, which should mean the only possible value for that type is `null`, but I'm guessing a lot of tooling conflates "nullable" with "optional" and we might need to handle this in the future.

### DIFF
--- a/packages/openapi-generator/src/openapi.ts
+++ b/packages/openapi-generator/src/openapi.ts
@@ -1,4 +1,4 @@
-import { OpenAPIV3_1 } from 'openapi-types';
+import { OpenAPIV3 } from 'openapi-types';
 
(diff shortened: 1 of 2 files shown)


bitgopatmcl:
Rebased to fix semantic merge conflict with  #537

andrew-scott-fischer (approved):

PR #537, opened 2023-09-05
fix: fix format of request bodies in openapi-generator

bitgopatmcl:
Caught this while trying to render a generated spec in SwaggerUI. This conflicts with #536 because of the `version` field, so depending on which one is merged first, the other will need to be fixed up.


ericcrosson-bitgo (approved):
Adds a `content` property under `requestBody`

PR #573, opened 2023-09-26
fix(superagent-wrapper): add @types/superagent as a peerDependency

andrew-scott-fischer:
@api-ts/superagent-wrapper requires types to be supplied based on your version of superagent


bitgopatmcl:
I opened https://github.com/BitGo/api-ts/pull/576 as an alternative since it seems wrong that the `@types` package should be a required dependency if this was used in a plain JS project or something.

bitgopatmcl:
Closing since this should be resolved by #576

PR #576, opened 2023-09-27
fix: remove import of type from `superagent`

bitgopatmcl:
Opening as an alternative to #573


ghost (approved):

andrew-scott-fischer:
Ready to merge?

bitgopatmcl:
Yeah just created a small local typescript project and it appears to build with this as a dependency and not `@types/superagent`

PR #581, opened 2023-10-03
Fix type declarations for io-ts 2.1.3 and upgrade Typescript

bitgopatmcl:
Allows Typescript to be ugraded to the latest version by fixing an issue with the types in `io-ts` 2.1.3. This is mostly a backport of the fix done in https://github.com/gcanti/io-ts/pull/657.

Rather than maintaining a fork of `io-ts`, I opted to instead export the fixed declarations into a vendored `@types/io-ts` package, then added `paths` entries in `tsconfig.json` to force `tsc` to favor these declarations over the ones bundled with upstream `io-ts`. This technique should be possible to replicate in other packages that want to upgrade TS without moving off of `io-ts` 2.1.3

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "pre-commit": "1.2.2",
         "prettier": "3.0.3",
         "turbo": "1.10.14",
-        "typescript": "4.6.4"
+        "typescript": "5.2.2"
       },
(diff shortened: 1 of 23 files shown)


bitgopatmcl:
Closing in favor of #582

PR #584, opened 2023-10-04
chore: update `flake.lock`

bitgopatmcl:
Bumps the node version to `v18.18.0` and gets rid of bad engine warnings on `npm install`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,12 +1,15 @@
 {
   "nodes": {
     "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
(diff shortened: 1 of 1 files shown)


flake.lock
```diff
@@ -1,12 +1,15 @@
 {
   "nodes": {
     "flake-utils": {
```
ericcrosson-bitgo, line 3: hmm, we're still using flake-utils? we can pickle that

ericcrosson-bitgo (approved):
Updates nix flake.lock

PR #619, opened 2023-11-08
chore: add build and test scripts to the root directory

ericcrosson-bitgo:
When I enter the repository directory, I habitually try to build and test
in that directory. Presently, that isn't supported in this repo.

This commit adds top-level build and test npm scripts so those actions
are supported.


bitgopatmcl (approved):
Moves workspace build and test scripts into `package.json`

PR #638, opened 2023-12-11
Add more missing `io-ts` codec definitions

bitgopatmcl:
Adds `t.any`, `t.unknown`, and `t.UnknownRecord`

### DIFF
--- a/packages/openapi-generator/src/knownImports.ts
+++ b/packages/openapi-generator/src/knownImports.ts
@@ -31,6 +31,8 @@ export const KNOWN_IMPORTS: KnownImports = {
     boolean: () => E.right({ type: 'boolean' }),
     null: () => E.right({ type: 'null' }),
     undefined: () => E.right({ type: 'undefined' }),
+    unknown: () => E.right({ type: 'any' }),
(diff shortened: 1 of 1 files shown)


andrew-scott-fischer (approved):

packages/openapi-generator/src/knownImports.ts
```diff
@@ -31,6 +31,8 @@ export const KNOWN_IMPORTS: KnownImports = {
     boolean: () => E.right({ type: 'boolean' }),
     null: () => E.right({ type: 'null' }),
     undefined: () => E.right({ type: 'undefined' }),
+    unknown: () => E.right({ type: 'any' }),
```
ericcrosson-bitgo, line 34: Cries in type-strictness :sob: 

Am I understanding correctly that the `unknown` in the LHS refers to TypeScript's `unknown` type, while the `any` on the RHS corresponds to OpenAPI's type?

If so, then I guess that's not as egregious as if both were TypeScript types!

bitgopatmcl: Yeah, the RHS is what eventually gets turned into an OpenAPI schema.

PR #640, opened 2023-12-21
Correctly parse comments in source files with leading comments

bitgopatmcl:
The result from `swc.parse` includes a `span` which is used to slice source files to pass to the comment parser. This number starts at an arbitrary value and continuously increments as source files are parsed. Unfortunately, `module.span.start` is the offset of the first parsed statement in a source file, ignoring any comments at the top of a file (so the first declaration's `span.start` == `module.span.start`). For the purposes of using these spans to slice the source for comments, we want to offset from the beginning of the actual file, not the first statement. Fortunately, this appears to be doable by tracking the previous span end.

### DIFF
--- a/packages/openapi-generator/test/codec.test.ts
+++ b/packages/openapi-generator/test/codec.test.ts
@@ -481,7 +481,7 @@ testCase('declaration comment is parsed', DECLARATION_COMMENT, {
       tags: [],
       source: [
         {
-          number: 2,
+          number: 0,
           source: '/**',
           tokens: {
             start: '',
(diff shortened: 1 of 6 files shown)


packages/openapi-generator/test/codec.test.ts
```diff
@@ -481,7 +481,7 @@ testCase('declaration comment is parsed', DECLARATION_COMMENT, {
       tags: [],
       source: [
         {
-          number: 2,
+          number: 0,
```
bitgopatmcl, line 484: Not sure what happened here since this is effectively a snapshot test, but `0` actually feels more correct

andrew-scott-fischer (approved):

PR #662, opened 2024-01-22
Handle `Object.assign` in openapi-generator

bitgopatmcl:
This is a slightly hacky way to do it, and will break if `Object.assign` is aliased, but that seems like a sufficiently rare edge case that it can be addressed if and when it comes up.


andrew-scott-fischer (approved):
Objects now `assign`able.

PR #680, opened 2024-02-16
feat(router): add support for request/response transformers

jonball4:
[feat(router): add support for request/response transformers](https://github.com/BitGo/api-ts/pull/680/commits/3a54035a1a0430422d3dbaadff0484f87398f54f)

### DIFF
--- a/packages/typed-express-router/src/errors.ts
+++ b/packages/typed-express-router/src/errors.ts
@@ -20,3 +20,21 @@ export function defaultOnEncodeError(
   res.status(500).end();
   console.warn(`Error in route handler: ${err}`);
 }
+
+export function defaultOnRequestTransformError(
+  err: unknown,
+  _req: express.Request,
+  res: express.Response,
+) {
+  res.status(500).end();
+  console.warn(`Error in route handler: ${err}`);
+}
+
+export function defaultOnResponseTransformError(
+  err: unknown,
+  _req: express.Request,
+  res: express.Response,
+) {
+  res.status(500).end();
+  console.warn(`Error in route handler: ${err}`);
+}
--- a/packages/typed-express-router/src/types.ts
+++ b/packages/typed-express-router/src/types.ts
@@ -1,4 +1,4 @@
-import { ApiSpec, HttpRoute, Method as HttpMethod } from '@api-ts/io-ts-http';
+import { ApiSpec, Method as HttpMethod, HttpRoute } from '@api-ts/io-ts-http';
 import express from 'express';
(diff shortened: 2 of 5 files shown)


stevenharrisBG (changes requested):
Please add jsdoc and update README
It really isn't at all clear how to use these changes without deep diving.

packages/typed-express-router/src/types.ts
```diff
@@ -37,25 +54,125 @@ export type AfterEncodedResponseSentFn = (
...
 };
 
+export type WrappedRouteOptions = UncheckedWrappedRouteOptions &
+  TransformedRouteOptions & { onDecodeError?: OnDecodeErrorFn };
+
 export type WrappedRouterOptions = express.RouterOptions & WrappedRouteOptions;
 
+type TypeTransformer<D, U> = (decoded: D) => U;
+
+type ApiMethodRequest<
```
manav-c, line 105:
```suggestion
type ApiRequest<
```

packages/typed-express-router/src/types.ts
```diff
@@ -37,25 +54,125 @@ export type AfterEncodedResponseSentFn = (
...
+  Spec extends ApiSpec,
+  ApiName extends keyof Spec,
+  Method extends keyof Spec[ApiName] & HttpMethod,
+> = NonNullable<Spec[ApiName][Method]>['request'];
+
+type ApiMethodRequestType<
```
manav-c, line 111:
```suggestion
type ApiRequestType<
```

packages/typed-express-router/src/errors.ts
```diff
@@ -20,3 +20,21 @@ export function defaultOnEncodeError(
   res.status(500).end();
   console.warn(`Error in route handler: ${err}`);
 }
+
+export function defaultOnRequestTransformError(
+  err: unknown,
+  _req: express.Request,
+  res: express.Response,
+) {
+  res.status(500).end();
```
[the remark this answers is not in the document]
manav-c: should update the default response error handler as well

manav-c:
Discussed with Jon on Slack - summary of my thoughts on these changes: These changes would make it clear to devs that there is a pattern of transforms that stems from thinking about the distinction between layers, like data and service -- but the transformation functions can easily be abused (just applying transformations in the route handler to meet the expectations of the transformation) or ignored completely. While these changes prompt devs think about transforms, I feel like the distinction pattern is more of a mindset that we need to cultivate within the team and am not sure if this PR helps with that specifically.

PR #684, opened 2024-02-22
Test: created test case for known line break error

ekorenblum-simtlix:
The openapi-generator cli will not recognize a comment block containing metadata if it's not written right after an empty line. More info [in this PR](https://github.com/BitGo/entity-validation-types/pull/185)


ekorenblum-simtlix:
Moving to PR #686

PR #686, opened 2024-02-23
Fix: Comment parsing was broken if the comment is not after a line break

ekorenblum-simtlix:
DX-222

The openapi-generator cli would not recognize a comment block containing metadata if it's not written right after an empty line.

More info [in this PR](https://github.com/BitGo/entity-validation-types/pull/185)

The extracted comment block would sometimes not include the opening slash, which broke parsing. Fixed by manually adding the character if not present

### DIFF
--- a/packages/openapi-generator/src/comments.ts
+++ b/packages/openapi-generator/src/comments.ts
@@ -6,6 +6,11 @@ export function leadingComment(
   start: number,
   end: number,
 ): Block[] {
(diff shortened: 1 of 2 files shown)


ericcrosson-bitgo (approved):
Parses JSDoc comments even when the usual newline is absent

@bitgopatmcl just putting this PR on your radar

PR #689, opened 2024-02-27
Fix: leadingComment function handles incorrectly sliced JSDoc blocks

ekorenblum-simtlix:
DX-223

The comment parser would previously break when swc's span positions weren't precise enough, thus some metadata would be lost when generating the swagger file

image


ericcrosson-bitgo (approved):
Works around the jank parsing!

Also cc @bitgopatmcl, to put this one on your radar. I know you have been talking about structural improvements to the span parsing

PR #700, opened 2024-03-08
fix: bump OpenAPI spec version to 3.0.3

ekorenblum-simtlix:
I meant to trigger a release for PR #669, trigger it with this commit instead.


ericcrosson-bitgo (approved):
Thanks @ekorenblum-simtlix!

PR #702, opened 2024-03-12
fix: generate a new release

ericcrosson-bitgo:
This is a no-op change whose motivation is to coerce
multi-semantic-release to publish a new release.


ericcrosson-bitgo:
image


https://nodejs.org/api/assert.html#strict-assertion-mode

bitgopatmcl (approved):

PR #725, opened 2024-03-30
Update nix flake impl

andrew-scott-fischer:
This commit removes npm based pre-commit hooks in favor of `pre-commit-hooks` managed by Nix.


flake.nix
```diff
@@ -1,34 +1,58 @@
...
-            '';
-          };
-        }
-      )
-    )
-  );
+          # Write the node version into a version-controlled file so we
+          # can point the actions/setup-node GitHub Action to use the
+          # same version.
+          node --version > .nvmrc
```
ericcrosson-bitgo, line 53:
```suggestion
```

.nvmrc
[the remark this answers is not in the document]
andrew-scott-fischer: I felt like this would enable someone who doesn't use nix to contribute, or at least build things locally. It also acts as a short hand to see which version of `node` a nix is using from the nixpkgs, since that's not outlined in the `flake.nix`.

Are you sure you prefer it to be removed?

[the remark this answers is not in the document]
ericcrosson-bitgo: > I felt like this would enable someone who doesn't use nix to contribute, or at least build things locally.

We should be ambivalent to which LTS version of Node.js a contributor uses, so this is not a large factor in my opinion.

> It also acts as a short hand to see which version of `node` nix is installing from the nixpkgs, since that's not outlined in the `flake.nix`.

We can simply run `node --version` to see what version of `node` Nix is installing, this isn't worth committing another file to our repository, especially one that is a derivation of other committed files.

> Are you sure you prefer it to be removed?

Yes please

ericcrosson-bitgo, line 1: Reminder: please remove this file too

andrew-scott-fischer: removed

ericcrosson-bitgo (approved):
Replaces pre-commit and lint-staged with pre-commit-hooks.nix :snowflake: :rocket: 

Thanks @andrew-scott-fischer!

PR #728, opened 2024-04-01
refactor: rename `*.yml` to `*.yaml`

andrew-scott-fischer:
I have been a fan of 3-letter extensions, but apparently the official recommended filename extnsion for YAML files has been `.yaml` since 2006[^1].

This commit renames files with a `.yml` extension to `.yaml`

[^1]: https://en.wikipedia.org/wiki/YAML#:~:text=The%20official%20recommended%20filename%20extension,yaml%20since%202006.


ericcrosson-bitgo (approved):
s/yml/yaml

PR #729, opened 2024-04-01
Track `unstable` tag in jsdoc

andrew-scott-fischer:
Introduces an `--unstable` cli flag that when set emits routes tagged with `@unstable` when generating OpenAPI specs from API specs.

This feature mirrors the functionality of the `--internal` cli flag identically, accommodating the orthogonal concern of generating OpenAPI specs which omit or include routes with an unstable interface.


bitgopatmcl (approved):
Adds support for `@unstable`

PR #730, opened 2024-04-01
fix: remove unused cli flags

andrew-scott-fischer:
The `--internal` and `--unstable` cli flags were never actually used by openapi-generator. This commit removes no functionality of the tool, it only cleans the cli interface.

Ticket: VL-1855

### DIFF
--- a/packages/openapi-generator/src/cli.ts
+++ b/packages/openapi-generator/src/cli.ts
@@ -1,15 +1,6 @@
 #!/usr/bin/env node
 
-import {
-  command,
-  run,
-  option,
-  string,
-  optional,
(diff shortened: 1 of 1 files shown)


bitgopatmcl (dismissed):
I believe that the `includeInternal` flag was in the original v1 of `openapi-generator`, and then was never implemented for `v2`.

Removes unused command line flags to enshrine existing behavior

bitgopatmcl (commented):
I can't suggest an edit in that location, but the `flag` and `boolean` imports are now unused

bitgopatmcl (approved):

PR #731, opened 2024-04-02
Add support for specifying base URL

bitgopatmcl:
Adds support for emitting the `servers` section of an OpenAPI spec by annotating `h.apiSpec` objects with the appropriate JSDoc tags. For example:

```javascript
    /** An API spec
     *
     * @url /api/v1
     **/
    export const test = h.apiSpec({
      'api.test': {
        get: h.httpRoute({
          path: '/test',
          method: 'GET',
          request: h.httpRequest({}),
          response: {
            200: t.string,
          },
        })
      }
    });
```

Will result in something like this in the emitted OpenAPI:

```yaml
servers:
- description: An API spec
  url: /api/v1
```

### DIFF
--- a/packages/openapi-generator/test/apiSpec.test.ts
+++ b/packages/openapi-generator/test/apiSpec.test.ts
@@ -4,7 +4,7 @@ import test from 'node:test';
 import type { NestedDirectoryJSON } from 'memfs';
 
 import { TestProject } from './testProject';
-import { parseApiSpec, type Route } from '../src';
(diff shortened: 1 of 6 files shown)


packages/openapi-generator/test/apiSpec.test.ts
ericcrosson-bitgo, line 1: It looks like we're trying to support a use case of specifying multiple servers. Should we add a test for that use-case?

ericcrosson-bitgo (approved):

PR #753, opened 2024-04-30
feat(openapi-generator): take and parse @example tags on JSDoc

ekorenblum-simtlix:
The OpenAPI generator module will now accept `@example` tags in the JSDoc, which will translate into OpenAPI examples. 
They may be used in the route definition to declare a response example, or in a schema definition to declare a schema usage example

### Schema definition

Given the following JSDoc schema definition 
```javascript
/**
 * A variable with example
 * 
 * @example foo
 */
```

The resulting YAML declaration will be
```yaml
summary: A variable with example
tags:
  example:
    test: foo
```

### Endpoint definition

Given the following endpoint definition

```javascript
import * as t from 'io-ts';
import * as h from '@api-ts/io-ts-http';

/**
 * A simple route
 *
 * @operationId api.v1.test
 * @tag Test Routes
 * @example {
 *   "test": "bar"
 * }
 */
export const route = h.httpRoute({
  path: '/foo',
  method: 'GET',
  request: h.httpRequest({}),
  response: {
    200: {
      test: t.string
    }
  },
});
```

The resulting YAML declaration will be
```yaml
openapi: 3.0.3
info:
  title: Test
  version: 1.0.0
paths:
  "/foo":
    get:
      summary: A simple route
      operationId: api.v1.test
      tags:
      - Test Routes
      parameters: []
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                properties:
                  test:
                    type: string
                required:
                - test
              example:
                test: bar
components:
  schemas: {}
```

### DIFF
--- a/packages/openapi-generator/src/jsdoc.ts
+++ b/packages/openapi-generator/src/jsdoc.ts
@@ -3,26 +3,49 @@ import type { Block } from 'comment-parser';
 export type JSDoc = {
   summary?: string;
   description?: string;
-  tags?: Record<string, string>;
+  tags?: Record<string, string> & { example?: unknown };
 };
 
 export function parseCommentBlock(comment: Block): JSDoc {
   let summary: string = '';
   let description: string = '';
-  let tags: Record<string, string> = {};
+  let tags: Record<string, string> & { example?: any } = {};
+  let writingExample = false;
 
   for (const line of comment.source) {
-    if (summary.length === 0) {
-      if (line.tokens.description === '') {
-        continue;
+    if (writingExample) {
+      tags['example'] = `${tags['example']}\n${line.source.split('*')[1]?.trim()}`;
+      try {
+        tags['example'] = JSON.parse(tags['example']);
+        writingExample = false;
+      } catch (e) {
+        if (line.source.endsWith('*/'))
+          throw new Error('@example contains invalid JSON');
+        else continue;
       }
-      summary = line.tokens.description;
     } else {
-      if (line.tokens.tag !== undefined && line.tokens.tag.length > 0) {
-        tags[line.tokens.tag.slice(1)] =
-          `${line.tokens.name} ${line.tokens.description}`.trim();
+      if (summary.length === 0) {
+        if (line.tokens.description === '') {
+          continue;
+        }
+        summary = line.tokens.description;
       } else {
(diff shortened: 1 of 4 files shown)


packages/openapi-generator/src/jsdoc.ts
```diff
@@ -3,26 +3,49 @@ import type { Block } from 'comment-parser';
 export type JSDoc = {
   summary?: string;
   description?: string;
-  tags?: Record<string, string>;
+  tags?: Record<string, string> & { example?: unknown };
```
bitgopatmcl, line 6: Technically this is unsound since `example` is a possible key of the record type so the final type should be the intersection of `string & unknown` which is just `string`

image

However, the compiler does treat this in the intended manner of "example could be anything, all other properties are strings". If you want the type to be more accurate you could write `Record<Exclude<string, 'example'>, string> & { example?: unknown }` but it doesn't appear to make a difference in practice.

packages/openapi-generator/test/jsdoc.test.ts
```diff
@@ -152,3 +152,93 @@ test('comment with a summary, description, and a tag in the middle of the descri
...
+});
+
+test('parameter with a comment and an example object (multi-line)', () => {
+  const comment = `
+    /**
+     * A variable with example
+     * 
+     * @example {
+     *   "test": "foo"
+     * }
```
bitgopatmcl, line 202: Can we add a test for tags that come after a multi-line example?

bitgopatmcl (changes requested):
Please re-word the commit message to use `feat(openapi-generator)` rather than `feature(openapi-generator)` since the latter won't be picked up by semantic-release.

packages/openapi-generator/test/jsdoc.test.ts
ekorenblum-simtlix: Added!

packages/openapi-generator/src/jsdoc.ts
ekorenblum-simtlix: Good catch! I changed it in the declaration as requested

bitgopatmcl (approved):

bitgopatmcl (approved):

PR #762, opened 2024-05-09
Don't silently drop properties when processing an `httpRoute`

bitgopatmcl:
Fixes an issue where errors are swallowed while processing an `httpRoute`. Generally this happens while processing a request type, the error just gets swallowed and the `request` property is omitted from the resulting schema. This then errors out later on in the section of the codebase that expects a request to be there, causing this hard-to-debug error:

> Error parsing SomeApiSpec: Route must have a request

With this change, the _actual_ error is passed all the way through, making it much easier to see what went wrong.


ericcrosson-bitgo (dismissed):

ericcrosson-bitgo (approved):
👨‍🍳 🚀

PR #766, opened 2024-05-15
feat(openapi-generator): support types imported from `node_modules`

anshchaturvedi:
Ticket: DX-389

This PR aims to introduce support to parse `types` from libraries living inside `node_modules`. Previously `openapi-generator` only parsed `types` that were defined locally and this heavily interfered with creating an automated solution for generating OpenAPI specifications when API's changes are introduced.

This PR also includes a more sophisticated end-to-end testing strategy so that any implementations that change the logic of how we generate our documentation will be caught by our tests.


packages/openapi-generator/src/project.ts
```diff
@@ -13,10 +13,28 @@ export class Project {
...
+      '@bitgo-private/query-param-types',
+      '@bitgo-private/request-types',
+      '@bitgo-private/schema-decoding',
+      '@bitgo-private/travel-rule-types',
+      '@bitgo-private/wallet-platform-types',
+      '@bitgo/public-types',
+      '@bitgo-private/compliance-tx-monitoring-types',
+      '@bitgo-private/event-service-types',
+      '@bitgo-private/ims-type',
+    ];
```
anshchaturvedi, line 37: note: do this dynamically

packages/openapi-generator/src/project.ts
```diff
@@ -75,13 +98,46 @@ export class Project {
...
     try {
+      path = this.maybeModifyPath(path);
+
+      // express doesn't parse for some reason
+      if (path.includes('express')) return E.right('express');
+
       const result = resolve.sync(path, {
         basedir,
-        extensions: ['.ts', '.js'],
+        extensions: ['.ts', '.js', '.d.ts'],
```
anshchaturvedi, line 117: question: do we still need `.d.ts`

packages/openapi-generator/src/symbol.ts
```diff
@@ -236,6 +236,24 @@ export function parseTopLevelSymbols(
...
+          if (targetExpr && 'arguments' in targetExpr) {
+            const targetArgument = targetExpr.arguments;
+            if (targetArgument && 'value' in targetArgument[0]!.expression) {
+              if (targetArgument[0]!.expression.type === 'StringLiteral') {
+                symbols.exportStarFiles.push(targetArgument[0]!.expression.value);
+              }
+            }
+          }
+        }
+      }
```
anshchaturvedi, line 256: probably don't need this

packages/openapi-generator/src/project.ts
```diff
@@ -98,37 +82,59 @@ export class Project {
...
-    return path;
+  private resolvePath(path: string, basedir: string): E.Either<string, string> {
+    try {
+      const result = resolve.sync(path, {
+        basedir,
+        extensions: ['.ts', '.js', '.d.ts'],
+      });
+
+      return E.right(result);
+    } catch (e: any) {
```
ericcrosson-bitgo, line 93: suggestion: use `unknown` here instead of `any`! 

It's preferable in just about every case (except for crazy generics) because the compiler will force you to inspect the value before you use it

packages/openapi-generator/package.json
```diff
@@ -17,7 +17,8 @@
     "clean": "rm -rf -- dist",
     "format": "prettier --check .",
     "format:fix": "prettier --write .",
-    "test": "c8 --all --src src node --require @swc-node/register --test test/*.test.ts"
+    "test": "c8 --all --src src node --require @swc-node/register --test test/*.test.ts",
+    "test:target": "c8 --all --src src node --require @swc-node/register"
```
anshchaturvedi, line 21: allows us to run specific test files

packages/openapi-generator/test/project.test.ts
```diff
@@ -93,3 +93,29 @@ testCase(
...
+  };
+
+  const project = new TestProject({ '/index.ts': NON_STRICT_MODE_SRC }, {});
+  await project.parseEntryPoint('/index.ts');
+  const sourceFile = project.get('/index.ts');
+
+  assert.strictEqual(sourceFile, undefined);
+  assert.strictEqual(errorCalled, true, new Error('console.error was not called'));
+
+  console.error = originalConsoleError;
```
anshchaturvedi, line 120: i wanted to make sure that we were indeed `console.error`'ing but I couldn't find a way to use `node:test` and catch the logs being sent to `stderr`, so this is the next best thing that I came up with. Verified that if you don't log to `stderr` the test fails. Lmk if there's a better way for this

packages/openapi-generator/src/project.ts
ad-world: resolved!

.prettierignore
```diff
@@ -1,3 +1,4 @@
 coverage/
 dist/
 flake.lock
+packages/openapi-generator/test/
```
ericcrosson-bitgo, line 4: Question: why exclude these files from prettier?

ad-world: That folder has some *intentionally* buggy code so that we can test errors and prettier complained about the buggy files, so as a quick and harmless workaround I added them here. I can look into another solution if you don't think we should ignore them though

ericcrosson-bitgo: Ah I see -- thanks for the context!

packages/openapi-generator/test/externalModule.test.ts
```diff
@@ -5,6 +5,7 @@ import * as p from 'path';
 
 import { parsePlainInitializer, Project, type Schema } from '../src';
 import { KNOWN_IMPORTS } from '../src/knownImports';
+// import { resolve } from 'node:path';
```
bitgopatmcl, line 8: nit: leftover commented out import

packages/openapi-generator/package.json
```diff
@@ -34,7 +35,7 @@
     "@types/resolve": "1.20.6",
     "c8": "9.1.0",
     "memfs": "4.9.2",
-    "typescript": "4.7.4"
+    "typescript": "^4.7.4"
```
bitgopatmcl, line 38: What necessitated this change? I think we can support this but would need to incorporate the patch [here](https://github.com/BitGo/api-ts/pull/582) (or a version of io-ts with that patch upstreamed to 2.1.3)

packages/openapi-generator/test/.prettierignore
```diff
@@ -0,0 +1 @@
+sample-types/
```
bitgopatmcl, line 1: Is this redundant with this entry? https://github.com/anshchaturvedi

bitgopatmcl (commented):
I'm wondering if this should count as a breaking change. Even though I haven't ran into it in testing, a package that previously didn't expect to have node_modules parsed may suddenly break upon upgrading if there is something that now causes an error in node_modules.

anshchaturvedi:
> I'm wondering if this should count as a breaking change. Even though I haven't ran into it in testing, a package that previously didn't expect to have node_modules parsed may suddenly break upon upgrading if there is something that now causes an error in node_modules.

I gave myself the task of running all the current `-types` repos that use `openapi-generator` so far and I don't recall seeing any errors, but yeah I guess it _could_ still break. Is there something special we do for releases that introduce breaking changes?

packages/openapi-generator/package.json
anshchaturvedi: Yeah I don't think we need this either, must've been something I introduced when trying to get everything working locally. We can remove this 👍🏽

packages/openapi-generator/test/.prettierignore
anshchaturvedi: hmm yeah I agree, will check this and remove as necessary.

bitgopatmcl:
> I gave myself the task of running all the current `-types` repos that use `openapi-generator` so far and I don't recall seeing any errors, but yeah I guess it _could_ still break. Is there something special we do for releases that introduce breaking changes?

The way to do with with semantic-release is to add a line in a commit body containing `BREAKING CHANGE: some description here`. Since this PR has a bunch of commits, it only needs to go in one of them somewhere.

anshchaturvedi:
> > I gave myself the task of running all the current `-types` repos that use `openapi-generator` so far and I don't recall seeing any errors, but yeah I guess it _could_ still break. Is there something special we do for releases that introduce breaking changes?
> 
> The way to do with with semantic-release is to add a line in a commit body containing `BREAKING CHANGE: some description here`. Since this PR has a bunch of commits, it only needs to go in one of them somewhere.

done! 17d919a7ea20df6696d22a365baa54e1c40a7f49

bitgopatmcl (approved):

PR #780, opened 2024-06-03
test: fix `openapi.test.ts` to properly test `@example` tag behaviour

anshchaturvedi:
The test I wrote before didn't test `@example` behaviour properly, this fixes it.

### DIFF
--- a/packages/openapi-generator/test/openapi.test.ts
+++ b/packages/openapi-generator/test/openapi.test.ts
@@ -2054,7 +2054,7 @@ export const route = h.httpRoute({
     query: {
       /** 
        * This is a bar param.
-       * @example { "foo": "bar" }
+       * @example "{ 'foo': 'bar' }"
       */
       bar: t.record(t.string, t.string),
     },
@@ -2103,6 +2103,7 @@ testCase('route with descriptions, patterns, and examples', ROUTE_WITH_DESCRIPTI
             required: true,
             schema: {
               type: 'object',
+              example: "{ 'foo': 'bar' }",
               additionalProperties: {
                 type: 'string'
               }


packages/openapi-generator/test/openapi.test.ts
```diff
@@ -2103,6 +2103,7 @@ testCase('route with descriptions, patterns, and examples', ROUTE_WITH_DESCRIPTI
             required: true,
             schema: {
               type: 'object',
+              example: "{ 'foo': 'bar' }",
```
bitgopatmcl, line 2106: if this is meant to be a JSON string shouldn't it contain escaped double quotes for keys and values instead of single ones?

anshchaturvedi: inside `ROUTE_WITH_DESCRIPTIONS_PATTERNS_EXAMPLES`, if I make it any one of

`@example '{ "foo": "bar" }'` or 
`@example "{ "foo": "bar" }"` or
`@example "{ \"foo\": \"bar\" }"`

then it doesn't seem to parse the tag properly (it probably has to do with the quotes because it cuts part of example and puts in the `description` section of the schema instead).

The only way it cooperates is how it is now. It suspect it has something to do with the `parseComment` function inside `comments.ts` but that isn't our implementation. Do you have any ideas of how we can work around this? Or maybe just leave it alone 🤔

bitgopatmcl: Ah, we probably should refactor that to not use the external `comment-parser` anyway

bitgopatmcl (approved):
Adds example tag test case

PR #787, opened 2024-06-06
feat: add support for `format` tag in openapi generator

ad-world:
DX-455

### DIFF
--- a/packages/openapi-generator/src/openapi.ts
+++ b/packages/openapi-generator/src/openapi.ts
@@ -139,6 +139,7 @@ function schemaToOpenAPI(
     const maxLength = getTagName(schema, 'maxLength');
     const minLength = getTagName(schema, 'minLength');
     const pattern = getTagName(schema, 'pattern');
+    const format = getTagName(schema, 'format');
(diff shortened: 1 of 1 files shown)


packages/openapi-generator/src/openapi.ts
```diff
@@ -151,6 +152,7 @@ function schemaToOpenAPI(
       ...(maxLength ? { maxLength: Number(maxLength) } : {}),
       ...(minLength ? { minLength: Number(minLength) } : {}),
       ...(pattern ? { pattern } : {}),
+      ...(format ? { format: `<${format}>` } : {}),
```
bitgopatmcl, line 155: Is this correct or should it just be a string as shown for example [here](https://stackoverflow.com/questions/77108530/modifying-the-format-field-in-springdoc-openapi-schema)?

ad-world: Oh you're right I was looking at dev-portal when i made this change. Will fix it

bitgopatmcl (approved):

PR #792, opened 2024-06-10
feat: added support for referencing existing specs in dev-portal

ekorenblum-simtlix:
DX-468

Many manually written API specs take advantage of existing schemas found under the `/content/shared` directory of `dev-portal`; the object declarations in `knownImports.ts` could be used for the `openapi-generator` tool to take advantage of them. This PR introduces that functionality, and adds `DateFromISOString` as its first supported schema


ekorenblum-simtlix:
It was decided against adding this feature in order to keep `openapi-generator`'s output completely independent from outside schemas.

PR #807, opened 2024-06-19
feat: add support for comments from declarations

anshchaturvedi:
In `wallet-platform-types` there are a lot of types (like `walletId` and `coin`) which are used in a LOT of files. If you want to add a `description` to that field then you need to copy-paste the comment for EVERY instance. But, if you try factoring out the fields to a common schema file and then export it where needed, openapi-generator ignores any comments and/or tags that you add to that declaration. We need to add support for this to allow for the refactors. 

This PR fixes that and adds a test for it.

Ticket: DX-494

### DIFF
--- a/packages/openapi-generator/src/cli.ts
+++ b/packages/openapi-generator/src/cli.ts
@@ -177,7 +177,7 @@ const app = command({
           );
           process.exit(1);
         }
-        const [newSourceFile, init] = initE.right;
+        const [newSourceFile, init, comment] = initE.right;
 
         const codecE = parseCodecInitializer(project.right, newSourceFile, init);
         if (E.isLeft(codecE)) {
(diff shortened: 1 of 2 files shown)


packages/openapi-generator/src/cli.ts
```diff
@@ -186,6 +186,10 @@ const app = command({
           );
           process.exit(1);
         }
+        if (comment !== undefined) {
+          codecE.right.comment = comment;
+        }
+
```
anshchaturvedi, line 192: This takes inspiration from `apiSpec.ts`:
https://github.com/BitGo/api-ts/blob/dd18f4663fecfb4533ecbe8935a66af1632857b4/packages/openapi-generator/src/apiSpec.ts#L75

bitgopatmcl (approved):

PR #812, opened 2024-06-26
fix: add `comment` block to `intersection` types

anshchaturvedi:
Ticket: DX-521

### DIFF
--- a/packages/openapi-generator/src/optimize.ts
+++ b/packages/openapi-generator/src/optimize.ts
@@ -28,7 +28,7 @@ export function foldIntersection(schema: Schema, optimize: OptimizeFn): Schema {
     }
   });
 
-  return schema.comment ? { ...result, comment: schema.comment } : result;
+  return result;
 }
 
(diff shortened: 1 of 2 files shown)


packages/openapi-generator/src/optimize.ts
```diff
@@ -28,7 +28,7 @@ export function foldIntersection(schema: Schema, optimize: OptimizeFn): Schema {
     }
   });
 
-  return result;
+  return schema.comment ? { ...result, comment: schema.comment } : result;
```
ad-world, line 31: I think this should be put into the main `optimize` function like it is for unions

ad-world: i.e. inside the big if statement, something like this

```
else if (schema.type === 'intersection') {
      const folded = foldIntersection(schema, optimize);
      if (schema.comment) { return { ...folded, comment: schema.comment }
      return folded
}
```

bitgopatmcl (approved):

ad-world (approved):

PR #819, opened 2024-07-05
feat: allow for `JSDoc`'s in `request` declarations

anshchaturvedi:
For some routes with union types in the request body, if there were no `@title` tags in the source repo definition of the route, then `dev-portal` would fallback to using the type (which was object). Trying to add these `@title` tags to declarations in `openapi-generator` was also not working since it wasn’t supported. 

Not cool:
image

Very cool:
image

Shoutout to @ad-world for finding this issue!

Ticket: DX-551

### DIFF
--- a/packages/openapi-generator/src/openapi.ts
+++ b/packages/openapi-generator/src/openapi.ts
@@ -260,25 +260,13 @@ function routeToOpenAPI(route: Route): [string, string, OpenAPIV3.OperationObjec
     {},
   );
 
-  const stripTopLevelComment = (schema: Schema) => {
-    const copy = { ...schema };
-
-    if (copy.comment?.description !== undefined && copy.comment?.description !== '') {
-      copy.comment.description = '';
-    }
-
-    return copy;
-  };
-
-  const topLevelStripped = stripTopLevelComment(route.body!);
-
   const requestBody =
     route.body === undefined
       ? {}
       : {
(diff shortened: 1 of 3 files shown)


packages/openapi-generator/src/openapi.ts
```diff
@@ -260,25 +260,13 @@ function routeToOpenAPI(route: Route): [string, string, OpenAPIV3.OperationObjec
...
-
-    if (copy.comment?.description !== undefined && copy.comment?.description !== '') {
-      copy.comment.description = '';
-    }
-
-    return copy;
-  };
-
-  const topLevelStripped = stripTopLevelComment(route.body!);
-
```
anshchaturvedi, line 274: this code is no longer needed after https://github.com/BitGo/api-ts/pull/810 was merged.

packages/openapi-generator/test/openapi.test.ts
```diff
@@ -28,8 +29,8 @@ async function testCase(
     if (sourceFile === undefined) {
       throw new Error('Failed to parse source file');
     }
-
-    const project = new Project();
+    const files: Record<string, SourceFile> = { './index.ts': sourceFile };
```
anshchaturvedi, line 32: so that the test harness can resolve `ref`'s

bitgopatmcl (approved):

PR #824, opened 2024-07-11
feat: rework metadata to use more openapi tags

ad-world:
DX-554

This PR adds known metadata into some of the known imports for the openapi-generator. i.e., it adds the `minLength: 1` tag for the `NonEmptyString` import, which gives readers of that type more better information on what kind of data that codec will accept.

### DIFF
--- a/packages/openapi-generator/src/knownImports.ts
+++ b/packages/openapi-generator/src/knownImports.ts
@@ -139,110 +139,183 @@ export const KNOWN_IMPORTS: KnownImports = {
   },
   'io-ts-numbers': {
     NumberFromString: () =>
-      E.right({ type: 'string', format: 'number', title: 'Number String' }),
+      E.right({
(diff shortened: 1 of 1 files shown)


packages/openapi-generator/src/knownImports.ts
ericcrosson-bitgo, line 1: @ad-world The PR body indicates this commit adds data, but it looks like it also removes the `title` property. Is that intentional?

ad-world: Yep - after a discussion with @vmccarty and @ekorenblum-simtlix, we decided that it would be best to represent these known imports using openapi tags that help describe the format of the data instead of just titles. We agreed that the titles didn't provide much value, where as using these tags to describe length, min, max, etc provide much more value

bitgopatmcl (approved):

PR #826, opened 2024-07-11
feat: add support for array examples with the @arrayExample tag

ad-world:


anshchaturvedi (commented):
nice addition! How does this render on `dev-portal`? Or do we still need to implement the front-end logic for this?

ad-world:
@anshchaturvedi since this is a custom tag, this isn't yet supported in dev-portal. I'd imagine its a pretty easy addition though!

anshchaturvedi (approved):
looks good!

PR #832, opened 2024-07-16
fix(deps): update whatwg-url to v14.0.0

ericcrosson-bitgo:
whatwg-url v14 drops support for Node.js 16 and Node.js 14, which may
be notable. Since these are not current LTS versions, I'm inclined to not
consider this a breaking change from the api-ts side.

Closes Ticket: VL-497


ericcrosson-bitgo:
whatwg-url changelog: https://github.com/jsdom/whatwg-url/releases

bitgopatmcl (approved):

PR #839, opened 2024-07-18
feat: support `minItem` and `maxItem` for array types

anshchaturvedi:
See example in ticket which inspired this PR.

Ticket: DX-602

### DIFF
--- a/packages/openapi-generator/src/openapi.ts
+++ b/packages/openapi-generator/src/openapi.ts
@@ -55,10 +55,12 @@ function schemaToOpenAPI(
           return undefined;
         }
 
-        const { arrayExample, ...rest } = defaultOpenAPIObject;
+        const { arrayExample, minItems, maxItems, ...rest } = defaultOpenAPIObject;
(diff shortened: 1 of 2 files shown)


packages/openapi-generator/src/openapi.ts
```diff
@@ -55,10 +55,12 @@ function schemaToOpenAPI(
           return undefined;
         }
 
-        const { arrayExample, ...rest } = defaultOpenAPIObject;
+        const { arrayExample, minItems, maxItems, ...rest } = defaultOpenAPIObject;
 
         return {
           type: 'array',
+          ...(minItems ? { minItems } : {}),
+          ...(maxItems ? { maxItems } : {}),
```
anshchaturvedi, line 63: there's probably more types we could support but this is a quick fix, as we encounter more types we want to support on the _outer_ level of the array we can reconsider this approach 👍🏽

ad-world (approved):
Looks good!

PR #842, opened 2024-07-23
feat: allow unions to consolidate into single primitive types

anshchaturvedi:
This PR aims to introduce a union-consolidation step, where we can simplify union schemas if they meet certain conditions.

**Motivation:**

We have encountered multiple codecs that type the same `A` type with different `I` types. For example, a value intended as a `boolean` at runtime might be passed into the API as a `boolean` (`true` or `false`), a `string` (`"true"` or `"false"`), or even a `number` (`0` or `1`). Currently, `dev-portal` displays all these options as valid ways to pass a boolean to the API. However, this can be overwhelming and unnecessary for clients. If the API ultimately expects a `boolean` (typed as `t.boolean`), we should simplify this by presenting only the primitive `boolean` type.

**Solution:**
We propose the following criteria for consolidating union schemas into a single primitive type:

1. The input (`I`) type is a union.
2. The runtime (`A`) type is a primitive type (specifically, `string`, `number`, or `boolean`).
3. The `A` type exists _literally_ within the `I` type.

To support this, we need to address two aspects that are not currently available:

1. **Identifying Primitive Types:**
   Add a `primitive` property to the `schema` type to identify primitive `io-ts` codecs.

2. **Determining Intermediate Types:**
   Add a `decodedType` property (name can be revised) to the `schema` type. This will specify the runtime type of codecs like `BooleanFromString` or `NumberFromString`, enabling us to determine if a union containing these codecs can be consolidated into a single primitive type.

**Ticket:** DX-614

### DIFF
--- a/packages/openapi-generator/src/ir.ts
+++ b/packages/openapi-generator/src/ir.ts
@@ -82,4 +82,9 @@ export type SchemaMetadata = Omit<
   | 'externalDocs'
 >;
 
-export type Schema = BaseSchema & HasComment & SchemaMetadata;
+type ExtendedSchemaMetadata = SchemaMetadata & {
+  primitive?: boolean;
+  decodedType?: string;
+};
+
+export type Schema = BaseSchema & HasComment & ExtendedSchemaMetadata;
--- a/packages/openapi-generator/src/knownImports.ts
+++ b/packages/openapi-generator/src/knownImports.ts
@@ -46,10 +46,10 @@ export const KNOWN_IMPORTS: KnownImports = {
     },
   },
   'io-ts': {
-    string: () => E.right({ type: 'string' }),
-    number: () => E.right({ type: 'number' }),
+    string: () => E.right({ type: 'string', primitive: true }),
+    number: () => E.right({ type: 'number', primitive: true }),
     bigint: () => E.right({ type: 'number' }),
-    boolean: () => E.right({ type: 'boolean' }),
+    boolean: () => E.right({ type: 'boolean', primitive: true }),
     null: () => E.right({ type: 'null' }),
     nullType: () => E.right({ type: 'null' }),
     undefined: () => E.right({ type: 'undefined' }),
@@ -143,11 +143,13 @@ export const KNOWN_IMPORTS: KnownImports = {
         type: 'string',
         format: 'number',
         pattern: '^\\d+$',
+        decodedType: 'number',
       }),
     NaturalFromString: () =>
       E.right({
         type: 'string',
         format: 'number',
+        decodedType: 'number',
       }),
     Negative: () =>
       E.right({
@@ -161,6 +163,7 @@ export const KNOWN_IMPORTS: KnownImports = {
         format: 'number',
         maximum: 0,
         exclusiveMaximum: true,
+        decodedType: 'number',
       }),
     NegativeInt: () =>
       E.right({
@@ -174,6 +177,7 @@ export const KNOWN_IMPORTS: KnownImports = {
         format: 'number',
         maximum: 0,
         exclusiveMaximum: true,
+        decodedType: 'number',
       }),
     NonNegative: () =>
       E.right({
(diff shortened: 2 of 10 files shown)


packages/openapi-generator/src/ir.ts
```diff
@@ -82,4 +82,9 @@ export type SchemaMetadata = Omit<
   | 'externalDocs'
 >;
 
-export type Schema = BaseSchema & HasComment & SchemaMetadata;
+type ExtendedSchemaMetadata = SchemaMetadata & {
+  primitive?: boolean;
+  intermediate?: string;
```
anshchaturvedi, line 87: this addition breaks a bunch of tests (note that these tests failing **do not signify that the generator doesn't work**, its just that some of our unit tests verify the intermediate schemas before they are converted to OpenAPI specs, and those fail because we don't have the `primitive` and `intermediate` fields in them).

**Question:** I'm an Elixir fanboy, and I know that in Elixir you can somehow get `virtual` fields (fields that exist but are not shown when printed or converted to JSON, but can be accessed like a normal field), is there a way to do that in TypeScript? I don't think so, but I thought I'd ask.

Anyways, fixing the tests is a non-issue, its just adding these fields to them. The only concern I see here really here is extra lines added to the tests.

packages/openapi-generator/src/knownImports.ts
anshchaturvedi, line 1: are there concerns with adding fields here?

packages/openapi-generator/src/optimize.ts
```diff
@@ -31,10 +31,49 @@ export function foldIntersection(schema: Schema, optimize: OptimizeFn): Schema {
...
+
+  /**
+   * We need to check three things:
+   * 1. All the schemas satisfy isIntermediateOrPrimitive
+   * 2. All the schemas have the same intermediate type (aka type at runtime, or the `A` type of the codec)
+   * 3. At least one of the schemas is a primitive type
+   *
+   * If all these conditions are satisfied, we can prove to ourselves that this is a union that
+   * we can consolidate to the intermediate (runtime) type.
+   */
```
anshchaturvedi, line 57: In my head this logic makes sense, but if anything looks amiss here lmk

packages/openapi-generator/src/ir.ts
ericcrosson-bitgo: Summarizing notes from our call:

1. I'm not aware of anything like Elixir's `virtual` fields, to that sounds pretty nifty
2. `intermediate` seems like an overloaded term, especially since we have an intermediate representation. I pondered a few alternative names like "runtime" or "decoded" type. I personally lean towards calling `A` the "decoded type", unless you can come up with something still clearer
3. I'm okay with adding extra lines to the other tests

packages/openapi-generator/test/optimize.test.ts
```diff
@@ -144,3 +144,39 @@ test('comments are exposed in objects', () => {
...
+
+  assert.deepEqual(optimize(input), expected);
+});
+
+test('non-consolidatable unions are not consolidated', () => {
+  const input: Schema = {
+    type: 'union',
+    schemas: [
+      { type: 'string', enum: ['true', 'false'], decodedType: 'boolean' },
+      { type: 'string', primitive: true },
```
anshchaturvedi, line 168: ideally we would remove `primitive` when `optimize`-ing this schema, but we need it for optimizing the `union` types, so we keep it here (as it gets discarded when the OpenAPI spec is created anyways)

packages/openapi-generator/src/ir.ts
```diff
@@ -82,4 +82,9 @@ export type SchemaMetadata = Omit<
   | 'externalDocs'
 >;
 
-export type Schema = BaseSchema & HasComment & SchemaMetadata;
+type ExtendedSchemaMetadata = SchemaMetadata & {
```
ad-world, line 85: nit: maybe rename this `SchemaUtil` or something else descriptive that doesn't include `Metadata` since metadata is used to describe the OpenAPI tags, where as these fields are used for something else!

ad-world (approved):

bitgopatmcl (approved):

PR #853, opened 2024-07-29
feat: support array examples at arbitrary levels

ad-world:
DX-622

This PR removes support for the @arrayExample jsdoc, and extends the functionality of the @example tag so that it can be used with arrays to show array example and non-array examples as well. 

This is how to use the new functionality.

Case 1. 
Want to show an example of an item within the array
```
/**
@example exampleItem
*/
field: t.array(t.string)
```

This will render as this:
```
{
    type: 'array',
    items: {
        type: 'string',
        example: 'exampleItem'
    }
}
```

Case 2. Want to render an array example
```
/**
@example ["exampleItem1", "exampleItem2"]
*/
field: t.array(t.string)
```

This is how it would render:
```
{
    type: 'array',
    example: ['exampleItem1', 'exampleItem2']
    items: {
        type: 'string'    
    }
}
```

Case 3. Want to render an item + array example

```
/**
@example btc
*/
const coinString = t.string

... 
/** 
@example ["btc", "eth"]
field: t.array(coinString)
*/
```

This is how it would render:
```
{
    type: 'array',
    example: ['btc', 'eth']
    items: {
        type: 'string',
        example: 'btc'
    }
}
```

### DIFF
--- a/packages/openapi-generator/src/openapi.ts
+++ b/packages/openapi-generator/src/openapi.ts
@@ -7,10 +7,6 @@ import type { Route } from './route';
 import type { Schema } from './ir';
 import { Block } from 'comment-parser';
 
-type ExtendedOpenApiSchema = OpenAPIV3.SchemaObject & {
-  arrayExample?: string;
-};
-
 function schemaToOpenAPI(
   schema: Schema,
 ): OpenAPIV3.SchemaObject | OpenAPIV3.ReferenceObject | undefined {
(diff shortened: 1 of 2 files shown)


ericcrosson-bitgo (commented):
Test cases look good.

What is your plan regarding call-sites that are using `arrayExample` today? 

Please add the missing semicolon that the linter flagged and once we have a solid roll-out plan, this looks good to merge

ad-world:
> What is your plan regarding call-sites that are using `arrayExample` today?

First step is to bump openapi-generator in dev-portal (after this is merged).
Next step is to update all @arrayExample's in WPT. There are only 2, so it is a trivial task.

ad-world:
https://github.com/BitGo/wallet-platform-types/pull/478 !

ericcrosson-bitgo (approved):
Seems fine to not consider this a breaking change since the removed `arrayExample` tag is undocumented and was recently added -- so adoption is expected to be minimal.

Approving but not merging yet so you can coordinate this with your other PRs

PR #855, opened 2024-07-30
feat: consolidate comments from nested levels of schemas

ad-world:
DX-592

This is a PR that adds functionality to consolidate comments when they are present in different schemas that refer to each other.

For example.

```typestrict
/** 
@example 123
*/
const EnterpriseIdString = t.string;

const Enterprise = t.type({
    /** The Enterprise ID */
    id: EnterpriseIdString
})
```

Previously, the example would get stripped from the generation since we saw the second comment and used it to overwrite the first one. This PR ensures that we don't lose that information.

### DIFF
--- a/packages/openapi-generator/src/comments.ts
+++ b/packages/openapi-generator/src/comments.ts
@@ -1,4 +1,5 @@
 import { parse as parseComment, Block } from 'comment-parser';
+import { Schema } from './ir';
 
 export function leadingComment(
   src: string,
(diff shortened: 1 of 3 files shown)


anshchaturvedi (dismissed):
We retain the top-level comment, re-adding it if it gets removed during the schema optimization step. We then construct a new CommentBlock that consolidates all the tags from the child schemas, ensuring that only the top-level tag is kept if duplicates are present.

The tests are comprehensive and look like they thoroughly cover the behavior. Everything looks great! Nice work! 🚀

bitgopatmcl (commented):
I like the approach here, but it needs to be rebased.

bitgopatmcl (approved):
:rocket: lgtm. I spent some time trying to think of a situation where we'd want to explicitly un-set a tag rather than overwrite it and am coming up blank

PR #856, opened 2024-07-31
feat: support private fields in the openapi generator

ad-world:
DX-613

This PR enables downstream users of api-ts to mark fields in their codecs with a `@private` jsdoc tag, which in turn will add the `x-internal: true` field within the respective schema for that field.

### DIFF
--- a/packages/openapi-generator/test/openapi.test.ts
+++ b/packages/openapi-generator/test/openapi.test.ts
@@ -3942,7 +3942,11 @@ import * as h from '@api-ts/io-ts-http';
 const SampleType = t.type({
   foo: t.string,
   /** @private */
-  bar: t.string, // This should show up with x-internal
+  bar: t.string, // This should show up with x-internal,
+  /** @private */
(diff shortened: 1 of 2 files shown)


bitgopatmcl (dismissed):

packages/openapi-generator/test/openapi.test.ts
```diff
@@ -17,11 +16,7 @@ import { SourceFile } from '../src/sourceFile';
 async function testCase(
   description: string,
   src: string,
-  expected: OpenAPIV3_1.Document<{
-    'x-internal'?: boolean;
-    'x-unstable'?: boolean;
-    'x-unknown-tags'?: object;
-  }>,
+  expected: any,
```
bitgopatmcl, line 19: You can probably do something like ```OpenAPIV3_1.Document<Record<`x-${string}`, any>>``` here, but it's a test function so probably not worth the hassle.

bitgopatmcl (approved):

PR #857, opened 2024-07-31
feat: support `domain` for `t.record` codec

anshchaturvedi:
### Problem:

Currently, `openapi-generator` only supports `t.record` with unrestricted domains. However, there can be codecs that use an `enumerable` (usually signified as a `t.keyof`) to specify that only certain keys are allowed and that all of these keys are required in the `t.record`.

Our `dev-portal` doesn't fully support rendering `t.record` that conforms to the [OpenAPI specification](https://swagger.io/docs/specification/data-models/dictionaries/), as the rendering code we ~copied~ adopted from [@stoplight/json-schema-viewer](https://github.com/stoplightio/json-schema-viewer) is outdated and a few versions behind the current release.

### Solution:

Although a `dev-portal` fix to render `t.record` better is in progress (though it won't be fully complete in its first iteration), we decided to circumvent the inadequate rendering by converting a `t.record` to a `t.type`. Both codecs expect all keys to exist in the object and all values of the keys to conform to the specified codec. With this PR, something like:
```typescript
export const SomeCodec = t.record(t.keyof({ foo: "foo", bar: "bar" }), t.type({ baz: t.string }))
```
will have its keys "expanded" and inlined by `openapi-generator` when generated, and will be equivalent to:
```typescript
export const SomeCodec = t.type({
  foo: t.type({ baz: t.string }),
  bar: t.type({ baz: t.string }),
})
```
>[!IMPORTANT]
>Any `t.record` codecs with an unrestricted domain (e.g., `t.string`) will remain unchanged.

Ticket: DX-637

### DIFF
--- a/packages/openapi-generator/src/openapi.ts
+++ b/packages/openapi-generator/src/openapi.ts
@@ -148,9 +148,24 @@ function schemaToOpenAPI(
         }
       case 'record':
         const additionalProperties = schemaToOpenAPI(schema.codomain);
-        if (additionalProperties === undefined) {
-          return undefined;
+        if (additionalProperties === undefined) return undefined;
+
+        if (schema.domain !== undefined) {
+          const keys = schemaToOpenAPI(schema.domain) as OpenAPIV3.SchemaObject;
+          if (keys.type === 'string' && keys.enum !== undefined) {
+            const properties = keys.enum.reduce((acc, key) => {
+              return { ...acc, [key]: additionalProperties };
+            }, {});
+
+            return {
+              type: 'object',
+              properties,
+              ...defaultOpenAPIObject,
+              required: keys.enum,
+            };
+          }
         }
+
         return {
           type: 'object',
           additionalProperties,
--- a/packages/openapi-generator/src/optimize.ts
+++ b/packages/openapi-generator/src/optimize.ts
@@ -226,14 +226,12 @@ export function optimize(schema: Schema): Schema {
     }
     return { type: 'array', items: optimized };
   } else if (schema.type === 'record') {
-    if (schema.comment) {
-      return {
-        type: 'record',
-        codomain: optimize(schema.codomain),
-        comment: schema.comment,
-      };
-    }
-    return { type: 'record', codomain: optimize(schema.codomain) };
+    return {
+      type: 'record',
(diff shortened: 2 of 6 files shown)


vmccarty (approved):

packages/openapi-generator/src/openapi.ts
```diff
@@ -148,9 +148,24 @@ function schemaToOpenAPI(
...
+            return {
+              type: 'object',
+              properties,
+              ...defaultOpenAPIObject,
+              required: keys.enum,
+            };
+          }
         }
+
         return {
```
ad-world, line 169: will the domain ever be undefined? i.e will we ever reach the bottom case?

ad-world: it looks like it's possible from the tests, just confused as to how we go from `t.string` as a key -> `undefined`

packages/openapi-generator/src/optimize.ts
```diff
@@ -226,14 +226,12 @@ export function optimize(schema: Schema): Schema {
...
-        codomain: optimize(schema.codomain),
-        comment: schema.comment,
-      };
-    }
-    return { type: 'record', codomain: optimize(schema.codomain) };
+    return {
+      type: 'record',
+      ...(schema.domain ? { domain: optimize(schema.domain) } : {}),
+      codomain: optimize(schema.codomain),
+      ...(schema.comment ? { comment: schema.comment } : {}),
```
ad-world, line 233: there MIGHT be a merge conflict here if my PR in api-ts gets merged first

packages/openapi-generator/src/openapi.ts
anshchaturvedi: yeah there are a couple codecs defined in `knownImports.ts` that don't define a `domain`, so in that case it would fall through to this case. See `JsonRecord` and `UnknownRecord`. Didn't want to change any existing behavior so kept the fallback case.

ad-world: oh i see, there another check when you do `keys.type === 'string' ...`. In this case, is the first check (`if (schema.domain !== undefined`) even necessary?

ad-world: sounds good!

ad-world (approved):
SUI 👍

PR #861, opened 2024-08-01
feat: consolidate unknown unions with well defined codecs

ad-world:
DX-644

This is a change to consolidate unknown unions in the openapi generator. For example, if we have something like this:

`t.union([WellDefinedCodec, t.unknown])` then it doesn’t make sense to pass through the t.unknown codec into the API docs, as we have our ‘best guess’ shape of the data present as well. This is a ticket to filter out these unknowns as they aren’t required, and add bloat to the generator’s output.


anshchaturvedi (approved):
I discussed this change with @ad-world. The goal is to onboard completely new codecs by unionizing them with the `t.unknown` type. This approach ensures that decoding/encoding operations do not fail and break the API if the codec is incorrect. It provides a better client experience because our `dev-portal` will avoid rendering an `any` schema, instead displaying a more informative and useful schema.

LGTM 🚀

PR #863, opened 2024-08-02
feat: add optional name param to httpRequest

ad-world:
DX-634

This ticket adds an optional name parameter to the httpRequest function. Now, if a decode fails (and the httpRequest has a name), the name can be used for logging / debugging purposes to improve inspection, rather than only seeing the expected schema and trying to figure out the request from there.

### DIFF
--- a/packages/io-ts-http/src/combinators.ts
+++ b/packages/io-ts-http/src/combinators.ts
@@ -1,4 +1,4 @@
-import { pipe } from 'fp-ts/pipeable';
+import { pipe } from 'fp-ts/function';
 import * as E from 'fp-ts/Either';
 import * as R from 'fp-ts/Record';
 import * as t from 'io-ts';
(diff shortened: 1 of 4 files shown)


bitgopatmcl (approved):

packages/io-ts-http/src/combinators.ts
```diff
@@ -1,4 +1,4 @@
-import { pipe } from 'fp-ts/pipeable';
+import { pipe } from 'fp-ts/function';
```
ericcrosson-bitgo, line 1: issue: This line is actually a breaking change! @ad-world, can you please open a PR to import from `pipeable` instead?

The issue is the io-ts-http package manifest advertises compatibility with fp-ts version 2.0.0 or higher[^1], while `pipe` was only added to the `function` module in fp-ts v2.6.3[^2]. It is important to import from `pipeable` (even though marked as deprecated) because that import does exist from v2.0.0[^3]. This guarantees even when consumers are using an fp-ts version between 2.0.0 and 2.6.3, the `pipe` import can still be found.

[^1]: https://github.com/BitGo/api-ts/blob/a7cd268553e3b068996596f4e09600fa913f700f/packages/io-ts-http/package.json#L23
[^2]: https://gcanti.github.io/fp-ts/modules/function.ts.html#pipe
[^3]: https://gcanti.github.io/fp-ts/modules/pipeable.ts.html#pipe

PR #869, opened 2024-08-07
refactor: introduce standardized error messages and stacktraces

anshchaturvedi:
This update adds `[INFO]`, `[WARN]`, and `[ERROR]` prefixes to error messages to make them easier to read and understand during debugging. It also includes stack traces for `E.left` returns outside of `cli.ts` to help with troubleshooting.

---
How logging looks:
image

How errors look:
image

Ticket: DX-660

### DIFF
--- a/packages/openapi-generator/src/cli.ts
+++ b/packages/openapi-generator/src/cli.ts
@@ -87,7 +87,7 @@ const app = command({
       const codecFilePath = p.resolve(codecFile);
       const codecModule = await import(codecFilePath);
       if (codecModule.default === undefined) {
-        console.error(`Could not find default export in ${codecFilePath}`);
+        console.error(`[ERROR] Could not find default export in ${codecFilePath}`);
         process.exit(1);
       }
       const customCodecs = codecModule.default(E);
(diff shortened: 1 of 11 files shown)


packages/openapi-generator/src/cli.ts
```diff
@@ -87,7 +87,7 @@ const app = command({
       const codecFilePath = p.resolve(codecFile);
       const codecModule = await import(codecFilePath);
       if (codecModule.default === undefined) {
-        console.error(`Could not find default export in ${codecFilePath}`);
+        console.error(`[ERROR] Could not find default export in ${codecFilePath}`);
```
ad-world, line 90: It might also make sense to convert `console.error([ERROR] ` into a function that takes an error message and formats it how you'd like, just so changing the format of all these error logs will be easier if we want to change that in the future.

i.e. `const logError = (msg: string) => console.error('[ERROR]' + msg);`

packages/openapi-generator/src/cli.ts
```diff
@@ -120,21 +120,21 @@ const app = command({
...
         console.error(
-          `Skipping ${symbol.name} because it is a ${symbol.init.callee.type}`,
+          `[WARN] Skipping ${symbol.name} because it is a ${symbol.init.callee.type}`,
         );
         continue;
       } else if (!isApiSpec(entryPoint, symbol.init.callee)) {
         continue;
       }
-      console.error(`Found API spec in ${symbol.name}`);
+      console.error(`[INFO] Found API spec in ${symbol.name}`);
```
ad-world, line 129: Log info function, etc

packages/openapi-generator/src/cli.ts
anshchaturvedi: good idea!

ad-world (approved):
Love it, this will be a very helpful change for downstream users!

SUI 👍

PR #870, opened 2024-08-08
feat: support custom codec files in host repositories

ad-world:
DX-658

This ticket adds support for types packages to define schemas for their own custom codecs (that are defined with `t.Type` or other un-supported primitives). 

To make use of this feature in the types package, first create a file titled `openapi-gen.config.js`. Then, add the following to the package.json in the package `"customCodecFile": "openapi-gen.config.js"`. Then, you can populate the file with your schemas like this:

```javascript
module.exports = E => {
    return { 
         CustomSchema: () => E.right({ type: 'string' })
    }
}
```

### DIFF
--- a/packages/openapi-generator/README.md
+++ b/packages/openapi-generator/README.md
@@ -5,7 +5,7 @@ API specification into an OpenAPI specification.
 
 ## Install
 
-```
+```shell
 npm install --save-dev @api-ts/openapi-generator
 ```
 
@@ -15,7 +15,7 @@ The **openapi-generator** assumes the io-ts-http `apiSpec` is exported in the to
 of the Typescript file passed as an input parameter. The OpenAPI specification will be
 written to stdout.
 
-```
+```shell
 ARGUMENTS:
   <file> - API route definition file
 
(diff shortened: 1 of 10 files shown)


packages/openapi-generator/src/project.ts
```diff
@@ -148,4 +157,40 @@ export class Project {
...
+        if (module.default === undefined) {
+          logError(`Could not find default export in ${customCodecPath}`);
+          return;
+        }
+
+        const customCodecs = module.default(E);
+        this.knownImports[packageName] = {
+          ...this.knownImports[packageName],
+          ...customCodecs,
+        };
```
ad-world, line 188: question for reviewers: do you think the order should be switched around here? i.e. `...customCodecs` comes first, so that if a downstream repo wants to redefine a schema for whatever reason, they can still do that rather than having it overwritten

ad-world: went ahead and made the change, let me know if you think it should be reverted. I think it makes sense to give the end user more priority.

vmccarty (dismissed):

packages/openapi-generator/README.md
```diff
@@ -35,32 +35,125 @@ For example:
...
+     };
+   };
+   ```
+
+2. The input parameter `E` is the namespace import of `fp-ts/Either`, which avoids
+   issues with `require`. The return type should be a `Record` containing AST
+   definitions for external libraries. For more information on the structure, refer to
+   [KNOWN_IMPORTS](./src/knownImports.ts).
+
+### Solution 2: Defining Custom Codec Schemas in the Types Package (recommended)
```
anshchaturvedi, line 93: if this is the recommended solution might as well make it `Solution #1` instead of `Solution #2`

packages/openapi-generator/src/cli.ts
anshchaturvedi, line 1: damn my `find and replace` failed me, thanks for fixing!

packages/openapi-generator/src/project.ts
```diff
@@ -148,4 +157,40 @@ export class Project {
...
+        const customCodecPath = resolve.sync(
+          `${packageName}/${packageInfo['customCodecFile']}`,
+          {
+            basedir,
+            extensions: ['.ts', '.js'],
+          },
+        );
+        const module = await import(customCodecPath);
+        if (module.default === undefined) {
+          logError(`Could not find default export in ${customCodecPath}`);
```
anshchaturvedi, line 180: This should be `errorLeft(`Could not find default export in ${customCodecPath}`)`. This approach allows the caller to handle (or propagate) the `E.left` up the call stack and manage the error. Since all logging is centralized in `cli.ts` for this project, we should maintain that practice.

packages/openapi-generator/src/project.ts
anshchaturvedi: I agree with this approach 👍🏽

packages/openapi-generator/src/project.ts
ad-world: This function doesn't return anything - calling errorLeft wouldn't do anything here

packages/openapi-generator/src/sourceFile.ts
```diff
@@ -41,7 +42,7 @@ export async function parseSource(
       span: module.span,
     };
   } catch (e: unknown) {
-    console.error(`Error parsing source file: ${path}`, e);
+    logError(`Error parsing source file: ${path}, ${e}`);
```
anshchaturvedi, line 45: we don't exit if we can't parse a sourcefile (just skip it instead), so maybe this should be a `logWarn` instead

packages/openapi-generator/src/project.ts
ad-world: We could rework this function to not have side effects and use Either for successes and failures

packages/openapi-generator/test/externalModuleApiSpec.test.ts
```diff
@@ -126,8 +126,8 @@ async function testCase(
       components,
     );
 
-    assert.deepEqual(errors, expectedErrors);
-    assert.deepEqual(openapi, expected);
+    assert.deepStrictEqual(errors, expectedErrors); 
+    assert.deepStrictEqual(openapi, expected);
```
anshchaturvedi, line 130: why do we need this change?

packages/openapi-generator/README.md
ad-world: done!

packages/openapi-generator/test/externalModuleApiSpec.test.ts
ad-world: https://nodejs.org/api/assert.html

It's not a required change, just ensures strict equality (which is what we want here)

packages/openapi-generator/src/project.ts
anshchaturvedi: I see, I think I misunderstood how this function is called. I'm okay with it (provided we change it to `logWarn` since it doesn't cause the program to exit), however I'm also open to using `Either` to represent failure/success. 

Though it might not be necessary to make this refactor if the caller isn't concerned with the result. I'll let you make the call here 👍🏽

ad-world: I think it makes sense to return an error here. It seems that in this case, the user is trying to write their own custom codecs but haven't configured the file properly. This should lead to an error

ad-world: I'll make the necessary changes

packages/openapi-generator/README.md
```diff
@@ -35,32 +35,125 @@ For example:
...
+By following these steps, the schemas for your custom codecs will be included in the
+generated API docs for any endpoints that use the respective codecs. The input parameter
+`E` is the namespace import of `fp-ts/Either`, and the return type should be a `Record`
+containing AST definitions for external libraries. For more details, see
+[KNOWN_IMPORTS](./src/knownImports.ts).
+
+### Solution 2: Using a Custom Codec Configuration File
+
+`openapi-generator` supports importing codecs from other packages in `node_modules`, but
+it struggles with `io-ts`primitives that need JavaScript interpretation, such as
```
anshchaturvedi, line 105:
```suggestion
it struggles with `io-ts` primitives that need JavaScript interpretation, such as
```

anshchaturvedi (approved):
🚀

PR #871, opened 2024-08-09
docs: rework openapi-generator README.md

ad-world:
DX-669

This is a ticket to rework the README for openapi-generator. This ticket adds a table of contents, and it also adds a generator reference at the end to showcase the different features that this generator supports for adding openapi documentation.

### DIFF
--- a/packages/openapi-generator/README.md
+++ b/packages/openapi-generator/README.md
@@ -3,13 +3,24 @@
 The api-ts **openapi-generator** is a command-line utility for converting an io-ts-http
 API specification into an OpenAPI specification.
 
-## Install
+## Contents
+
+1. [Installation](#1-install)
+2. [Usage](#2-use)
+3. [Preparing a types package for reusable codecs](#3-preparing-a-types-package-for-reusable-codecs)
+4. [Defining schemas for custom codecs](#4-defining-custom-codecs)
+5. [List of supported io-ts primitives](#5-list-of-supported-io-ts-primitives)
+6. [Generator Reference](#6-generator-reference)
+   1. [Endpoint documentation](#61-endpoint-documentation)
+   2. [Schema documentation](#62-schema-documentation)
+
+## 1. Install
 
 ```shell
 npm install --save-dev @api-ts/openapi-generator
 ```
 
-## Use
+## 2. Use
 
 The **openapi-generator** assumes the io-ts-http `apiSpec` is exported in the top level
 of the Typescript file passed as an input parameter. The OpenAPI specification will be
@@ -35,7 +46,7 @@ For example:
 npx openapi-generator src/index.ts
 ```
 
-## Preparing a types package for reusable codecs
+## 3. Preparing a types package for reusable codecs
 
 In order to use types from external `io-ts` types packages, you must ensure two things
 are done.
@@ -49,14 +60,14 @@ are done.
    the types in the source code. For example, if the entrypoint is `src/index.ts`, then
    set `"types": "src/index.ts"` in the `package.json`
 
-## Defining Custom Codecs
+## 4. Defining Custom Codecs
 
 When working with `openapi-generator`, you may encounter challenges with handling custom
 codecs that require JavaScript interpretation or aren't natively supported by the
 generator. These issues typically arise with codecs such as `new t.Type(...)` and other
 primitives that aren't directly supported. However, there are two solutions to address
-these challenges effectively. Click [here](#list-of-supported-io-ts-primitives) for the
-list of supported primitives.
+these challenges effectively. Click [here](#5-list-of-supported-io-ts-primitives) for
+the list of supported primitives.
 
 ### Solution 1: Defining Custom Codec Schemas in the Types Package (recommended)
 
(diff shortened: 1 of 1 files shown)


packages/openapi-generator/README.md
```diff
@@ -157,3 +168,320 @@ Follow these steps to create and use a custom codec configuration file:
...
 - UnknownRecord
 - void
+
+## 6. Generator Reference
+
+This section will highlight all the features that this generator supports, with examples
+to help you add meaningful documentation to your code that will allow clients to use our
+APIs with ease.
+
+### 1. Endpoint documentation
```
andrew-scott-fischer, line 178: Is this more consistent?

```suggestion
### 6.1 Endpoint documentation
```

packages/openapi-generator/README.md
```diff
@@ -157,3 +168,320 @@ Follow these steps to create and use a custom codec configuration file:
...
+
+Given an endpoint defined using `h.httpRoute`, you can add documentation and metadata to
+this endpoint through the use of jsdocs. Here are the following list of attributes that
+are supported.
+
+#### 1.1 Summary
```
andrew-scott-fischer, line 184:
```suggestion
#### 6.1.1 Summary
```

packages/openapi-generator/README.md
```diff
@@ -157,3 +168,320 @@ Follow these steps to create and use a custom codec configuration file:
...
+specification as the endpoints' summary
+
+```javascript
+/**
+ * This is the summary
+ */
+const route = h.httpRoute({ ... })
+```
+
+#### 1.2 Description
```
andrew-scott-fischer, line 196:
```suggestion
#### 6.1.2 Description
```

packages/openapi-generator/README.md
```diff
@@ -157,3 +168,320 @@ Follow these steps to create and use a custom codec configuration file:
...
+ * @version 3
+ */
+const route = h.httpRoute({ ... })
+```
+
+#### 1.9 Sample output
+
+This is what the OpenAPI specification will look like for the route we have built.
+
+```json
```
anshchaturvedi, line 333: haven't tested it but i think setting this to `javascript` should get rid of the weird red highlighting, since its not valid `json` but valid `javascript`...? (maybe im wrong lol). Or you could just make it valid JSON instead

packages/openapi-generator/README.md
```diff
@@ -131,7 +142,7 @@ Follow these steps to create and use a custom codec configuration file:
    definitions for external libraries. For more information on the structure, refer to
    [KNOWN_IMPORTS](./src/knownImports.ts).
 
-## List of supported io-ts primitives
+## 5. List of supported io-ts primitives
 
 - string
```
anshchaturvedi, line 147: nit for all the bullet points
```suggestion
- `string`
```

anshchaturvedi, line 147: nit for all the bullet points
```suggestion
- `string`
```

packages/openapi-generator/README.md
ad-world: yep, done this for 6.1, 6.2, and subsections

packages/openapi-generator/README.md
ad-world: made most of it valid json, the red highlighting doesn't show up in the rendered markdown btw

packages/openapi-generator/README.md
ad-world: dun

packages/openapi-generator/README.md
anshchaturvedi: image

ad-world: oh bruh

ad-world: it doesn't show up on vscode

ad-world: anyway fixed

packages/openapi-generator/README.md
```diff
@@ -131,29 +142,345 @@ Follow these steps to create and use a custom codec configuration file:
...
+            "key": "value"
+          }
+        },
+        "tag": "Wallet",
+        "x-internal": true, // @private
+        "x-unstable": true, // @unstable
+        "x-unknown-tags": {
+          "version": 3
+        }
+        ...parameters,
```
anshchaturvedi, line 353: nit: will fix the slightly-off highlighting
```suggestion
        },
        ...parameters,
```

anshchaturvedi (dismissed):
looks great! i didn't exactly proofread but did a quick skim and its very information, so that is always a good thing. good stuff Arya!

anshchaturvedi (dismissed):

bitgopatmcl (dismissed):

packages/openapi-generator/README.md
```diff
@@ -131,29 +142,345 @@ Follow these steps to create and use a custom codec configuration file:
...
+ * This is description line 1
+ * This is description line 2
+ *
+ * @unstable
+ * @operationId v2.sample.route
+ * @tag Wallet
+ */
+const route = h.httpRoute({ ... })
+```
+
```
andrew-scott-fischer, line 286: These docs are great! I know these are the `openapi-generator` docs, but do we have anywhere that indicates how these work with defining when a breaking change will occur to a `-types` library like `wallet-platform-types`? 

To add more color, a route tagged with the `@unstable` tag won't be considered as a part of the formal api, and `BREAKING CHANGE: `s to that route won't be considered as such. Do we have a place to document such behavior?

packages/openapi-generator/README.md
```diff
@@ -131,29 +142,345 @@ Follow these steps to create and use a custom codec configuration file:
...
+#### 6.2.3 Custom Tags
+
+These are some tags that you can use in your schema jsdocs are custom to this generator.
+
+- `@private` allows you to mark any field as in any schema as private. The final spec
+  will have `x-internal: true` for schemas with the `@private` tag.
+- `@deprecated` allows to mark any field in any schema as deprecated. The final spec
+  will include `deprecated: true` in the final specificaiton.
```
andrew-scott-fischer, line 474: Question: should we include `@unstable` here? It is a custom tag, but I'm not sure if a route tagged with `@unstable` has documentation generated for it or not.

ad-world: `@unstable` is only supported at the endpoint level, not lower schema/field level so I don't think it makes sense to add it here. I'm not even sure what support for that would look like lol

packages/openapi-generator/README.md
ad-world: I'm not sure I follow the behavior you are describing. `@unstable` routes won't be documented, so a breaking change to an unstable route wouldn't change any documentation anywhere (since the route was unstable to begin with).

packages/openapi-generator/README.md
andrew-scott-fischer: My mistake! Thank you.

ad-world:
@andrew-scott-fischer am I good to merge? There's another unresolved comment!

packages/openapi-generator/README.md
andrew-scott-fischer: When I implemented the `@unstable` feature, I think the expectation was that it would show up in docs, but be marked as "unstable" to indicate it was in a state of active development, and that breaking changes can occur. The idea was to make the idea of an "unstable" endpoint orthogonal from a "private" endpoint. If it was marked as "unstable", but not "private", that was something we wanted to make available to consumers, and thus should generate docs for.

andrew-scott-fischer:
@ad-world [I've left a comment](https://github.com/BitGo/api-ts/pull/871#discussion_r1714218869)

packages/openapi-generator/README.md
ad-world: I believe routes tagged @unstable are now stripped from the spec. I believe routes are marked as unstable when their service function definitions are incomplete, or haven't been migrated. It doesn't make sense to generate docs in that case, since the endpoint may not even exist yet in the typed router. So yes, that expectation has changed and it is reflected [here](https://github.com/BitGo/gha-dev-portal-updater/blob/3fcdb004f7eb4d345f7e16eb8df5867c011ebb6d/.github/workflows/generate-and-update.yml#L116).

andrew-scott-fischer: I understand that's what the code you've linked does, but I'm not sure that's what the customer would intend to do if they used the `@unstable` tag. My concern isn't blocking to your PR, but I raise it since we should favor having several functions that do one thing, then composing them together to get the desired effect. 

I feel like [the line of code you referenced](https://github.com/BitGo/gha-dev-portal-updater/blob/3fcdb004f7eb4d345f7e16eb8df5867c011ebb6d/.github/workflows/generate-and-update.yml#L116) overloads the `@unstable` functionality. If someone wants to generate docs for an `@unstable` endpoint, that is no longer possible. 

I know that `@private` means don't generate docs for the public dev-portal, but do generate them for the private dev-portal, so `@private` has it's own feature set that it represents. If the idea of "do not generate docs for this endpoint" is a specific feature we have, maybe there should be a `@ignore-doc-generation` tag (bike-shed the name) which encapsulates this functionality, rather than overloading `@unstable`.

packages/openapi-generator/README.md
```diff
@@ -131,29 +142,345 @@ Follow these steps to create and use a custom codec configuration file:
...
+ * @version 3
+ */
+const route = h.httpRoute({ ... })
+```
+
+#### 6.1.9 Sample output
+
+This is what the OpenAPI specification will look like for the route we have built.
+
+```javascript
```
ericcrosson-bitgo, line 333: praise: Clever, using the JavaScript language here to avoid the JSON syntax errors!

packages/openapi-generator/README.md
```diff
@@ -131,29 +142,345 @@ Follow these steps to create and use a custom codec configuration file:
...
+
+#### 6.1.8 Unknown tags
+
+Any other tags that are added to this top-level will be classified as an uknown tag, and
+will be placed inside the `x-unknown-tags` field in the OpenAPI specification. You can
+use this feature to write custom workflows and filtering logic for you full
+specification. For example, you could add a `@version` tag and have a workflow that
+filters endpoints based on the version field in the `x-unknown-tags` field.
+
+```javascript
```
ericcrosson-bitgo, line 314: For the code snippets, can you please use `typescript` consistently? This will aid copy/pasting and ensure if we add type data to a block, the syntax is displayed as intended

packages/openapi-generator/README.md
```diff
@@ -131,29 +142,345 @@ Follow these steps to create and use a custom codec configuration file:
...
+## 6. Generator Reference
+
+This section will highlight all the features that this generator supports, with examples
+to help you add meaningful documentation to your code that will allow clients to use our
+APIs with ease.
+
+### 6.1. Endpoint documentation
+
+Given an endpoint defined using `h.httpRoute`, you can add documentation and metadata to
+this endpoint through the use of jsdocs. Here are the following list of attributes that
```
ericcrosson-bitgo, line 181: issue: Will you please replace all instance of jsdoc with JSDoc? It's a proper noun and this is how they capitalize it on their [official docs](https://jsdoc.app/).

It's all right if that happens in a subsequent PR

packages/openapi-generator/README.md
ericcrosson-bitgo: I agree in general with the points @andrew-scott-fischer has raised.

Question: does openapi-generator strip unstable routes from the OpenAPI output, or does a different component in our build system do the stripping? Maybe we should migrate from the first implementation to the second, if that's how we want the feature to work with our internal API docs, but for this general-purpose tool, it seems sensible to have one feature for one intended outcome

I think this thread is worth keeping alive, but not necessarily blocking this PR since the addition is accurate for today's implementation

ericcrosson-bitgo (dismissed):
Thanks for the thorough review, all!

packages/openapi-generator/README.md
ad-world: Got it - this makes sense. Then I'll remove that line from the generic updater, and users already have the ability to create their own logic for removing endpoints (or whatever custom functionality they want to add at the endpoint level) described in section 6.1.8

packages/openapi-generator/README.md
ad-world: @anshchaturvedi is the 🐐 for recommending this

ad-world:
@ericcrosson-bitgo the actual stripping happens in another step of the process [here](https://github.com/BitGo/gha-dev-portal-updater/blob/3fcdb004f7eb4d345f7e16eb8df5867c011ebb6d/.github/workflows/generate-and-update.yml#L116). But I'm now thinking that we may just get rid of this, as @andrew-scott-fischer has mentioned that it is not the expected result of the @unstable tag to remove the docs completely, rather than annotate them as unstable. I guess we just need to make sure that this information is captured correctly in dev-portal

packages/openapi-generator/README.md
ad-world: Here is a PR to fix that: https://github.com/BitGo/gha-dev-portal-updater/pull/69/files

ad-world: Docs have also been updated to show the new effect of `@unstable`

ericcrosson-bitgo:
@ad-world Will you please squash this PR's commits?

anshchaturvedi (approved):

PR #882, opened 2024-08-23
feat: add type error if path doesn't start with /

ad-world:
DX-703

### DIFF
--- a/packages/io-ts-http/src/httpRoute.ts
+++ b/packages/io-ts-http/src/httpRoute.ts
@@ -13,8 +13,10 @@ export const Method = t.keyof({
 
 export type Method = t.TypeOf<typeof Method>;
 
+export type PathString = `/${string}`;
+
 export type HttpRoute<M extends Method = Method> = {
-  readonly path: string;
+  readonly path: PathString;
   readonly method: Uppercase<M>;
   readonly request: HttpRequestCodec<any>;
   readonly response: HttpResponse;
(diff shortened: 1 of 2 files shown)


packages/io-ts-http/src/httpRoute.ts
```diff
@@ -13,8 +13,10 @@ export const Method = t.keyof({
 
 export type Method = t.TypeOf<typeof Method>;
 
+export type PathString = `/${string}`;
```
bitgopatmcl, line 16: I'm torn between whether this counts as a breaking change, or something more akin to "this was always intended and the previous behavior is a bug". I'm curious to hear what others think about this.

ad-world: Would an endpoint work if the path didn't didn't start with a /?

bitgopatmcl: I can test with superagent and express, but right now am assuming it is sort of undefined behavior and a leading `/` may be implicitly added

bitgopatmcl: I suppose viewing that as undefined behavior is an argument against it being a breaking change.

ad-world: I'm also not opposed to marking this as a breaking change! I can add that to the commit msg if needed.

bitgopatmcl (approved):

PR #901, opened 2024-09-13
Revert "feat: add type error if path doesn't start with /"

bitgopatmcl:
Causes type errors when multiple copies of `io-ts-http` 3.x are in node_modules, which is unfortunately easy to do.

Fixes:
```
Types of property 'path' are incompatible.
                Type 'string' is not assignable to type '`/${string}`'.
```

This reverts commit a27e388397c48f4e3fdf35971aab595904223766.


ericcrosson-bitgo (approved):
Is this a commit we should ship in a v4 then?

PR #913, opened 2024-10-03
fix(superagent-wrapper): revert avoid mutation by sketchy codecs

jonball4:
This reverts commit d2ac27b929eb25215418a41a12f52381c2bd4c95.


domoneill18 (approved):

domoneill18:
For additional context, use of structured clone prevented upgrades in consumers of the package because it resulted in a DataCloneError when params passed had a method/methods nested in it's values.

bitgopatmcl (approved):

PR #925, opened 2024-10-22
Sort routes by path before writing the output JSON in openapi-generator

bitgopatmcl:
Sort routes by path before writing the output JSON in `openapi-generator`.

* Modify `convertRoutesToOpenAPI` function in `packages/openapi-generator/src/openapi.ts` to sort routes by path before constructing the `paths` object.
* Add test case in `packages/openapi-generator/test/openapi/base.test.ts` to verify routes are sorted by path in the output JSON.

BREAKING CHANGE: Changing the sort order may affect consumers that review diffs of the generated JSON

### DIFF
--- a/packages/openapi-generator/src/openapi.ts
+++ b/packages/openapi-generator/src/openapi.ts
@@ -431,11 +431,21 @@ export function convertRoutesToOpenAPI(
     {} as Record<string, OpenAPIV3.SchemaObject | OpenAPIV3.ReferenceObject>,
   );
 
+  const sortedPaths = Object.keys(paths)
+    .sort((a, b) => a.localeCompare(b))
+    .reduce(
+      (acc, key) => {
+        acc[key] = paths[key]!;
+        return acc;
+      },
+      {} as Record<string, OpenAPIV3.PathItemObject>,
+    );
+
   return {
     openapi: '3.0.3',
     info,
     ...(servers.length > 0 ? { servers } : {}),
-    paths,
+    paths: sortedPaths,
     components: {
       schemas: openapiSchemas,
     },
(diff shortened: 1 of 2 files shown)


packages/openapi-generator/src/openapi.ts
```diff
@@ -431,11 +431,21 @@ export function convertRoutesToOpenAPI(
     {} as Record<string, OpenAPIV3.SchemaObject | OpenAPIV3.ReferenceObject>,
   );
 
+  const sortedPaths = Object.keys(paths)
+    .sort((a, b) => a.localeCompare(b))
+    .reduce(
+      (acc, key) => {
+        acc[key] = paths[key]!;
```
bitgopatmcl, line 438: This works because insertion order [is preserved for string keys](https://stackoverflow.com/questions/69946683/can-i-preserve-the-order-of-a-javascript-objects-entries-when-some-entry-keys#comment123645050_69946683)

ericcrosson-bitgo: Hmm, yes I see this in the comment you linked:

> entries come in the order in which the properties were added

Followed immediately by:

> Relying on object property ordering in JavaScript is a really bad idea, because it makes code extremely fragile. If you need an ordering, create an array with the properties in it in the order that works for your application.

😅 

How do you feel about using a Map instead of an object? The overview of Map on [MDN] reads:

> The Map object holds key-value pairs and remembers the original insertion order of the keys.

[MDN]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_and_maps_compared

bitgopatmcl: I can try and see how it serializes.

bitgopatmcl: No luck:
> YAMLException: unacceptable kind of an object to dump [object Map]

ericcrosson-bitgo: All righty, well, so long as it's tested and deterministic!

ericcrosson-bitgo (approved):
Sorts route by path

PR #927, opened 2024-10-24
Sort HTTP verbs within each path in OpenAPI spec

bitgopatmcl:
Sort HTTP verbs within each path in the OpenAPI specification.

* Modify `convertRoutesToOpenAPI` function in `packages/openapi-generator/src/openapi.ts` to sort HTTP verbs within each path after sorting the paths.
* Add a test case in `packages/openapi-generator/test/openapi/base.test.ts` to verify the sorting of HTTP verbs within each path.


ericcrosson-bitgo (approved):
Sorts verbs within each path

PR #937, opened 2024-10-31
Support simple computed properties

bitgopatmcl:
Adds support for computed properties, but only simple ones that resolve to string or number literals.

### DIFF
--- a/packages/openapi-generator/src/codec.ts
+++ b/packages/openapi-generator/src/codec.ts
@@ -223,7 +223,8 @@ function parseObjectExpression(
     } else if (
       property.key.type !== 'Identifier' &&
       property.key.type !== 'StringLiteral' &&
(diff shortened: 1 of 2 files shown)


packages/openapi-generator/src/codec.ts
```diff
@@ -223,7 +223,8 @@ function parseObjectExpression(
     } else if (
       property.key.type !== 'Identifier' &&
```
jathu5, line 224: random nit: !['Identifier', 'StringLiteral', 'NumericLiteral', 'Computed'].includes(property.key.type)

jathu-oai (approved):

ericcrosson-bitgo (approved):

anshchaturvedi:
lets gooooooooooooooo

PR #988, opened 2025-01-16
feat: Prioritize 'source' property in package.json

bitgopatmcl:
Ticket: VL-2669


ericcrosson-bitgo (approved):
Prioritizes `source` in package.json, if exists

PR #1002, opened 2025-02-10
fix: implement lazy file processing with improved dependency handling

bitgopatmcl:
Makes file processing lazy, which should fix errors when a package coincidentally imports a non-io-ts related package that isn't actually used as part of the api schema.


ericcrosson-bitgo (approved):
Great feature!

PR #1004, opened 2025-02-11
feat: Add JSON parsing examples

youngjungithub:


website/docs/how-to-guides/parsing.json-strings.md
ericcrosson-bitgo, line 1: Thanks for the example, @youngjungithub! Can you put this in the context of api-ts? Looks like we have a test case using JsonFromString here:

https://github.com/BitGo/api-ts/blob/18d8b41dd5e2bf0d373b85014ea6d677acd38c20/packages/io-ts-http/test/httpRequest.test.ts#L49

Notably, removingn the `t.string` causes compilation errors on my machine.

youngjungithub: Hello! I am looking into `encodedJson: t.string.pipe(JsonFromString).pipe(t.type({ jsonKey: t.null })),`  and removing the `t.string` indeed causes a compilation error. I will investigate. Seems very intriguing.

andrew-scott-fischer (dismissed):

website/docs/how-to-guides/parsing.json-strings.md
```diff
@@ -0,0 +1,81 @@
+# Unpacking a Subtle Quirk in JsonFromString Codec
```
vmccarty, line 1: Can this title be more specific?

website/docs/how-to-guides/parsing.json-strings.md
```diff
@@ -0,0 +1,81 @@
...
+
+There's a small detail in `JsonFromString` from the `io-ts-types` that is worth noting.
+There are three type parameters to a codec `t.Type<I, O, A>`. The third parameter
+determines the type that the `decode` function receives. Most other codecs have the
+third parameter set to `unknown`. But `JsonFromString`'s type is
+`t.Type<Json, string, string>`. As a result, `JsonFromString` expects a string type
+before passing it to `decode`. You can easily convert `JsonFromString` to
+`t.Type<Json, string, unknown>` using `t.string`. See the example below:
```
vmccarty, line 9:
```suggestion
There are three type parameters to a codec: `t.Type<I, O, A>`. The third parameter determines the type that the `decode` function receives. Most codecs have the third parameter set to `unknown`. However, the type for `JsonFromString` in `io-ts-types` is `t.Type<Json, string, string>`. Therefore, `JsonFromString` expects a string type before passing it to `decode`. You can easily convert `JsonFromString` to `t.Type<Json, string, unknown>` using `t.string`. 

For example:
```

website/docs/intro.md
```diff
@@ -2,11 +2,16 @@
...
-```python hello.py
-# mark[16:24]
-print("This is Code Hike")
-```
+`io-ts-http` brings type safety to HTTP data handling in TypeScript by embracing the
+"[parse, don't validate]" philosophy. Rather than simply checking if incoming HTTP data
+is valid, it also parse raw, less-structured data (like strings or JSON) into strongly
+typed, precise objects using the io-ts library. This parsing happens at the system
+boundary, ensuring all types have use cases in your code. Once parsed, you can trust
+that the data satisfies both [type and semantic analysis].
```
vmccarty, line 12:
```suggestion
`io-ts-http` brings type safety to HTTP data handling in TypeScript by embracing the "[parse, don't validate]" philosophy. Rather than simply checking if incoming HTTP data is valid, it also parses raw, less-structured data (like strings or JSON) into strongly typed, precise objects using the io-ts library. This parsing happens at the system boundary, ensuring all types have use cases in your code. Once parsed, you can trust the data satisfies both [type and semantic analysis].
```

vmccarty (commented):
Not sure if this helps. I mostly just did some editing to the first body of text.

website/docs/intro.md
youngjungithub: suggestions implemented.

website/docs/how-to-guides/parsing.json-strings.md
vmccarty: I'd avoid using possessives for non-human nouns 👽 How about `Understanding Atypical Type Behavior from JsonFromString in io-ts-types`?

website/docs/how-to-guides/_category_.json
```diff
@@ -1,7 +1,8 @@
 {
-  "label": "How-To Guide",
+  "label": "How - To Guides",
```
vmccarty, line 2:
```suggestion
  "label": "How-To Guides",
```

youngjungithub: suggestions applied

youngjungithub:
@ericcrosson-bitgo Ready for review!

website/docs/how-to-guides/parsing.json-strings.md
```diff
@@ -0,0 +1,82 @@
+# Understanding Atypical Type Behavior from JsonFromString in io-ts-types
+
+There are three type parameters to a codec: `t.Type<I, O, A>`. The third parameter
+determines the type that the `decode` function receives. Most codecs have the third
+parameter set to `unknown`. However, the type for `JsonFromString` in `io-ts-types` is
+`t.Type<Json, string, string>`. Therefore, `JsonFromString` expects a string type before
+passing it to `decode`. You can easily convert `JsonFromString` to
+`t.Type<Json, string, unknown>` using `t.string`.
```
ericcrosson-bitgo, line 8: suggestion: It's not quite accurate to say we can _convert_ `JsonFromString` to another type; instead we're chaining codecs so the output of the first codec aligns with the expected `I` type of `JsonFromString`.

How about the following copy?


> Though we know headers, url parameters, and query parameters will be received as a `string` or `string[]` value, due to a limitation in api-ts, `httpRequest` only accepts codecs that decode values starting from the `unknown` type. 
> Consequently, decoding a header, url parameter, or query parameter with a codec like `JsonFromString`, which can only decode values typed as `string`, will produce a error like:
> 
> ```
> Type 'Type<Json, string, string>' is not assignable to type 'Mixed'.
>   Types of property 'validate' are incompatible.
>     Type 'Validate<string, Json>' is not assignable to type 'Validate<unknown, any>'.
>       Type 'unknown' is not assignable to type 'string'.
> ```
> 
> There's a straightforward pattern you can use when you have a value typed as `unknown` but need to decode it with a codec that can only decode a narrower type. This pattern is called _codec chaining_:
> 
> ```ts
> declare const JsonFromString: t.Type<Json, string, string>;
> declare const t.string: t.Type<string, string, unknown>;
> 
> const myCodec: t.Type<Json, string, unknown> = t.string.pipe(JsonFromString);
> ```
> 
> Here, `t.string` decodes a value from `unknown` to `string`, and then `JsonFromString` decodes the same value from `string` to `Json`.

I'm sure the flow and some word choices can be improved. What do you think?

vmccarty (dismissed):
LGTM

youngjungithub (commented):
Working on it!

website/docs/how-to-guides/parsing.json-strings.md
youngjungithub: @ericcrosson-bitgo @vmccarty Thank you for your insights. How does this sound?


Headers, URL parameters, and query parameters come as `string` or `string[]` values. But, due to a limit in api-ts, `httpRequest` only accepts codecs that decode from an `unknown` type. Consequently, decoding a header, URL parameter, or query parameter with a codec like `JsonFromString`, which can only decode values typed as `string`, will produce an error:
```
Type 'Type<Json, string, string>' is not assignable to type 'Mixed'.
  Types of property 'validate' are incompatible.
    Type 'Validate<string, Json>' is not assignable to type 'Validate<unknown, any>'.
      Type 'unknown' is not assignable to type 'string'.
```
You can follow a simple pattern when you have an `unknown` value. If you need to decode it, use a codec that works with a more specific type. We call this pattern codec chaining.
```
declare const JsonFromString: t.Type<Json, string, string>;
declare const t.string: t.Type<string, string, unknown>;

const myCodec: t.Type<Json, string, unknown> = t.string.pipe(JsonFromString);
```
`t.string` decodes a value from `unknown` to `string`. Then, `JsonFromString` decodes that `string` into `JSON`.

ericcrosson-bitgo (approved):
🚀 

Thank you, @youngjungithub!

PR #1014, opened 2025-03-02
fix: handle text responses

piotrfrankowski:


ericcrosson-bitgo (approved):
Thanks @piotrfrankowski!

ericcrosson-bitgo:
Whoops, I fat-fingered the rebase. Re-opened as #1047!

PR #1024, opened 2025-03-17
style: format openapi-generator test files

ericcrosson-bitgo:
I believe openapi-generator's test files were excluded from formatting
because back in the day, they used a different format to declare tests
and formatting was intrusive. That doesn't seem to apply today, so this
commit brings openapi-generator's tests back into the fold (of happily
formatted files).

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,3 @@
 coverage/
 dist/
 flake.lock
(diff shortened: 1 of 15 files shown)


bitgopatmcl (approved):
Formats a bunch of test files

PR #1026, opened 2025-03-18
docs: add Diátaxis taxonomy labels to documentation categories

youngjungithub:
[Ticket](https://bitgoinc.atlassian.net/browse/DX-1149)

- Update sidebars.js with explicit category definitions
- Modify how-to guides category description
- Create reference/explanation category scaffolding
- Align documentation structure with Diátaxis framework

### DIFF
--- a/website/docs/explanation/parsing-codec-chaining.md
+++ b/website/docs/explanation/parsing-codec-chaining.md
@@ -1,0 +1,102 @@
+# Decoding JSON from Headers, Query Parameters, and URL Parameters
+
+Though we know headers, URL parameters, and query parameters will be received as a
+`string` or `string[]` value, due to a limitation in api-ts, `httpRequest` only accepts
+codecs that decode values starting from the `unknown` type. Consequently, decoding a
+header, URL parameter, or query parameter with a codec like `JsonFromString`, which can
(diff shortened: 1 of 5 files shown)


website/docs/explanation/parsing-codec-chaining.md
ericcrosson-bitgo, line 1: @youngjungithub This duplicates content between the how-to guide and explanation. I've heard our technical writers mention how this can increase the maintenance burden, because now we have to keep two copies of the same data in sync over time.

Is this intentional, or do you have any ideas on how we can duplicate only the essential information?

youngjungithub: Yes it is intentional. To minimize maintenance burden we:

1. Keep procedural steps in How-To Guides
2. Store explanatory context in shared components
3. Use cross-references instead of duplications

ericcrosson-bitgo: How does that translate to a plan for this document? https://github.com/BitGo/api-ts/blob/audit-docs/website/docs/how-to-guides/parsing.json-strings.md

ericcrosson-bitgo (approved):

PR #1044, opened 2025-06-25
feat(opentelemetry): create otel instrumentation for typed-express-router

starfy84:
Ticket: DX-1473

This PR implements `opentelemetry` instrumentation for the `decode` and `sendEncoded` calls in `wrapRouter`. 

Changes:
- `typed-express-wrapper`
  - `@opentelemetry/api@^1.0.0` added as an optional peer dependency
  - `@opentelemetry/sdk-trace-base@1.30.1` , `@opentelemetry/sdk-trace-node@1.30.1`, and `@opentelemetry/api@1.9.0` added as dev dependencies
  - `wrapRouter` modified to create decode and encode spans if `@opentelemetry/api` is installed
    - if `@opentelemetry/api` is not installed, spans are not created
  - `onDecodeError` option removed. Please use `decodeErrorFormatter` and `getDecodeErrorStatusCode`
    - `decodeErrorFormatter` takes in an array of `ValidationError`s and a `WrappedRequest`, returning a `Json` object.
    - `getDecodeErrorStatusCode` takes in an array of `ValidationError`s and a `WrappedRequest`, returning a number.
  - `onEncodeError` option removed. Please use `encodeErrorFormatter` and `getEncodeErrorStatusCode`
    - `encodeErrorFormatter` takes in an error and a `WrappedRequest`, returning a `Json` object.
    - `getEncodeErrorStatusCode` takes in an error and a `WrappedRequest`, returning a number.
  - `typed-express-router` now handles the sending of the http response when there is a decode or encode error 
- `express-wrapper`
  - `routerForApiSpec` modified to pass in new parameters `decodeErrorFormatter`, `getDecodeErrorStatusCode`, `encodeErrorFormatter`, and `getEncodeErrorStatusCode`
    - These new parameters can also be customized by modifying the props passed into the function (see `CreateRouterProps`)

BREAKING CHANGE: `onDecodeError` and `onEncodeError` have been removed. Please use `decodeErrorFormatter`, `getDecodeErrorStatusCode`, `encodeErrorFormatter`, and `getEncodeErrorStatusCode`

### DIFF
--- a/packages/express-wrapper/src/request.ts
+++ b/packages/express-wrapper/src/request.ts
@@ -8,8 +8,10 @@ import * as PathReporter from 'io-ts/lib/PathReporter';
 
 import { ApiSpec, HttpRoute, RequestType, ResponseType } from '@api-ts/io-ts-http';
 import {
-  OnDecodeErrorFn,
-  OnEncodeErrorFn,
+  type DecodeErrorFormatterFn,
+  type EncodeErrorFormatterFn,
+  type GetDecodeErrorStatusCodeFn,
+  type GetEncodeErrorStatusCodeFn,
   TypedRequestHandler,
 } from '@api-ts/typed-express-router';
 
@@ -94,17 +96,28 @@ const createNamedFunction = <F extends (...args: any) => void>(
   fn: F,
 ): F => Object.defineProperty(fn, 'name', { value: name });
 
-export const onDecodeError: OnDecodeErrorFn = (errs, _req, res) => {
+export const defaultDecodeErrorFormatter: DecodeErrorFormatterFn = (errs, _req) => {
   const validationErrors = PathReporter.failure(errs);
-  const validationErrorMessage = validationErrors.join('\n');
-  res.writeHead(400, { 'Content-Type': 'application/json' });
-  res.write(JSON.stringify({ error: validationErrorMessage }));
-  res.end();
+  return { error: validationErrors.join('\n') };
 };
 
-export const onEncodeError: OnEncodeErrorFn = (err, _req, res) => {
+export const defaultEncodeErrorFormatter: EncodeErrorFormatterFn = (_err, _req) => {
+  return {};
+};
+
+export const defaultGetDecodeErrorStatusCode: GetDecodeErrorStatusCodeFn = (
+  _errs,
+  _req,
+) => {
+  return 400;
+};
+
+export const defaultGetEncodeErrorStatusCode: GetEncodeErrorStatusCodeFn = (
+  err,
+  _req,
+) => {
   console.warn('Error in route handler:', err);
-  res.status(500).end();
+  return 500;
 };
 
 export const handleRequest = (
--- a/packages/typed-express-router/src/errors.ts
+++ b/packages/typed-express-router/src/errors.ts
@@ -1,22 +1,22 @@
-import express from 'express';
-import { Errors } from 'io-ts';
 import * as PathReporter from 'io-ts/lib/PathReporter';
 
-export function defaultOnDecodeError(
-  errs: Errors,
-  _req: express.Request,
-  res: express.Response,
-) {
-  const validationErrors = PathReporter.failure(errs);
-  const validationErrorMessage = validationErrors.join('\n');
-  res.status(400).json({ error: validationErrorMessage }).end();
-}
+import type {
+  DecodeErrorFormatterFn,
+  EncodeErrorFormatterFn,
+  GetDecodeErrorStatusCodeFn,
+  GetEncodeErrorStatusCodeFn,
+} from './types';
 
-export function defaultOnEncodeError(
-  err: unknown,
-  _req: express.Request,
-  res: express.Response,
-) {
-  res.status(500).end();
-  console.warn(`Error in route handler: ${err}`);
-}
+export const defaultDecodeErrorFormatter: DecodeErrorFormatterFn = PathReporter.failure;
+
+export const defaultGetDecodeErrorStatusCode: GetDecodeErrorStatusCodeFn = (
+  _err,
+  _req,
+) => 400;
+
+export const defaultEncodeErrorFormatter: EncodeErrorFormatterFn = () => ({});
+
+export const defaultGetEncodeErrorStatusCode: GetEncodeErrorStatusCodeFn = (
+  _err,
+  _req,
+) => 500;
--- a/packages/typed-express-router/src/types.ts
+++ b/packages/typed-express-router/src/types.ts
@@ -2,6 +2,12 @@ import { ApiSpec, HttpRoute, Method as HttpMethod } from '@api-ts/io-ts-http';
 import express from 'express';
 import * as t from 'io-ts';
 
+export type Json = boolean | number | string | null | JsonArray | JsonRecord;
+export interface JsonRecord {
+  readonly [key: string]: Json;
+}
+export interface JsonArray extends ReadonlyArray<Json> {}
+
 export type Methods = Lowercase<HttpMethod>;
 
 export type WrappedRequest<Decoded = unknown> = express.Request & {
@@ -18,17 +24,9 @@ export type WrappedResponse<Responses extends {} = Record<string | number, unkno
     ) => void;
   };
 
-export type OnDecodeErrorFn = (
-  errs: t.Errors,
-  req: express.Request,
-  res: express.Response,
-) => void;
+export type EncodeErrorFormatterFn = (err: unknown, req: WrappedRequest) => Json;
 
-export type OnEncodeErrorFn = (
-  err: unknown,
-  req: WrappedRequest,
-  res: WrappedResponse,
-) => void;
+export type GetEncodeErrorStatusCodeFn = (err: unknown, req: WrappedRequest) => number;
 
 export type AfterEncodedResponseSentFn = (
   status: number,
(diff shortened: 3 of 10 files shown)


packages/express-wrapper/src/request.ts
```diff
@@ -97,9 +97,15 @@ const createNamedFunction = <F extends (...args: any) => void>(
 export const onDecodeError: OnDecodeErrorFn = (errs, _req, res) => {
   const validationErrors = PathReporter.failure(errs);
   const validationErrorMessage = validationErrors.join('\n');
-  res.writeHead(400, { 'Content-Type': 'application/json' });
-  res.write(JSON.stringify({ error: validationErrorMessage }));
-  res.end();
+  return {
```
starfy84, line 100: `express-wrapper` modified due to `onDecodeError` return type being changed

packages/typed-express-router/src/errors.ts
```diff
@@ -2,14 +2,19 @@ import express from 'express';
...
 import * as PathReporter from 'io-ts/lib/PathReporter';
 
+import { OnDecodeErrorFn } from './types';
+
 export function defaultOnDecodeError(
   errs: Errors,
   _req: express.Request,
   res: express.Response,
-) {
+): ReturnType<OnDecodeErrorFn> {
```
ericcrosson-bitgo, line 11: We can simplify this thing we wrote several hours into our pair programming session last week with:

```typescript
export const defaultOnDecodeError: OnDecodeErrorFn = (
  errs,
  _req, // Use _req to indicate it's not used if you prefer
  res
) => {
  // ... 
};
```

packages/typed-express-router/src/types.ts
```diff
@@ -44,16 +43,25 @@ export type AfterEncodedResponseSentFn = (
   res: WrappedResponse,
 ) => void;
 
+export type PathReporterFailureFn = (
+  es: Array<t.ValidationError>,
+  req: WrappedRequest,
+) => Json;
+
+export type GetDecodeErrorStatusCodeFn = (
+  es: Array<t.ValidationError>,
```
ericcrosson-bitgo, line 52: Suggestion: How do you feel about the name `errors` instead of `es`, which I assume meant "plural of e"? My concern is that it looks like a typo in the usual `req`/`res` combo!

packages/typed-express-router/src/errors.ts
```diff
@@ -1,22 +1,21 @@
...
-) {
-  res.status(500).end();
-  console.warn(`Error in route handler: ${err}`);
-}
+export const defaultGetDecodeErrorStatusCode: GetDecodeErrorStatusCodeFn = (
+  _err,
+  _req,
+) => 400;
+
+export const defaultEncodeErrorFormatter: EncodeErrorFormatterFn = () => ({});
```
ericcrosson-bitgo, line 16: idem

packages/typed-express-router/test/server.test.ts
```diff
@@ -538,8 +538,8 @@ test('should invoke custom encode error function when an unknown keyed status is
   });
 
   const router = createRouter(WeirdApi, {
-    onEncodeError: (_err, _req, res) => {
-      res.status(500).json('Custom encode error').end();
+    encodeErrorFormatter: (_err, _req) => {
+      return 'Custom encode error';
```
ericcrosson-bitgo, line 542: praise: Look at all that data over state! 😍

packages/typed-express-router/test/telemetry.test.ts
```diff
@@ -0,0 +1,465 @@
...
+} from '@opentelemetry/sdk-trace-base';
+import { NodeTracerProvider } from '@opentelemetry/sdk-trace-node';
+import { AsyncLocalStorageContextManager } from '@opentelemetry/context-async-hooks';
+import { Span, SpanKind, context, trace } from '@opentelemetry/api';
+import { setRPCMetadata, RPCType } from '@opentelemetry/core';
+import express from 'express';
+import { createRouter } from '../src';
+import assert from 'node:assert';
+import { ApiTsAttributes } from '../src/telemetry';
+import test, { beforeEach, afterEach } from 'node:test';
```
ericcrosson-bitgo, line 17: quibble: please separate imports into the following sections, with sections separated by a single newline:

1. Native node imports (begin with `node:`)
2. Imports from a named package
3. Relative imports

> Note: Not blocking change, can be done in a follow-up PR.

packages/typed-express-router/test/telemetry.test.ts
```diff
@@ -0,0 +1,465 @@
...
+
+const makeRequest = (
+  server: http.Server,
+  method: string,
+  path: string,
+  body?: any,
+): Promise<{ statusCode: number; data: string }> => {
+  return new Promise((resolve, reject) => {
+    const url = `http://localhost:${(server.address() as any).port}${path}`;
```
ericcrosson-bitgo, line 26: question: LSP says `server.address()` is of type `string | AddressInfo | null` -- how do you always know it is of type `AddressInfo` here? I wonder if there's a way to clean up the type cast.

packages/typed-express-router/test/telemetry.test.ts
```diff
@@ -0,0 +1,465 @@
...
+      id: t.string,
+    },
+  }),
+  response: {
+    200: t.type({
+      id: t.string,
+    }),
+  },
+});
+
```
ericcrosson-bitgo, line 93: quibble: Please use the conventional order of GET, POST, PUT, DELETE, PATCH, by moving this GET operation above the PUT operation declared for the same path on line 58.

https://github.com/BitGo/api-ts/blob/a2033285ea449797d945422c5f915a2375bfcb2b/packages/typed-express-router/src/index.ts#L190-L194

> Note: Not a blocking change, can be done in a follow-up PR.

packages/typed-express-router/test/telemetry.test.ts
```diff
@@ -0,0 +1,465 @@
...
+const TestApiSpec = apiSpec({
+  'hello.world': {
+    put: PutHello,
+    get: GetHello,
```
ericcrosson-bitgo, line 112: quibble: For the same reason as above, please use the order of GET, POST, PUT, DELETE, PATCH

> Note: Not a blocking change, can be done in a follow-up PR.

packages/typed-express-router/test/telemetry.test.ts
```diff
@@ -0,0 +1,465 @@
...
+    assert.ok(decodeSpan, 'Decode span not found');
+    assert.strictEqual(
+      decodeSpan?.attributes?.[ApiTsAttributes.API_TS_OPERATION_ID],
+      'hello.world',
+    );
+    assert.strictEqual(
+      decodeSpan?.attributes?.[ApiTsAttributes.API_TS_PATH],
+      '/hello/{id}',
+    );
+    assert.strictEqual(decodeSpan?.attributes?.[ApiTsAttributes.API_TS_METHOD], 'GET');
```
ericcrosson-bitgo, line 235: shorthand now as we are on a call: please update ordering to match the order keys are introduced in the                                          object definition

> Note: Not a blocking change, can be done in a follow-up PR.

packages/typed-express-router/test/telemetry.test.ts
```diff
@@ -0,0 +1,465 @@
...
+      'hello.world',
+    );
+    assert.strictEqual(encodeSpan?.attributes?.[ApiTsAttributes.API_TS_PATH], '/hello');
+    assert.strictEqual(encodeSpan?.attributes?.[ApiTsAttributes.API_TS_METHOD], 'PUT');
+  });
+});
+
+// TODO: POST
+// TODO: DELETE
+// TODO: PATCH
```
ericcrosson-bitgo, line 298: noting

packages/typed-express-router/test/telemetry.test.ts
```diff
@@ -0,0 +1,465 @@
...
+    });
+    rootSpan.end();
+
+    assert.strictEqual(response.statusCode, 400);
+
+    // Check spans
+    const spans = getSpans();
+
+    // Find encode span
+    const decodeSpan = spans.find(
```
ericcrosson-bitgo, line 361: noting irregular use of encode/decode

packages/typed-express-router/test/telemetry.test.ts
starfy84: According to the documentation: https://github.com/nodejs/node/blob/main/doc/api/net.md#serveraddress

> For a server listening on a pipe or Unix domain socket, the name is returned

and

> server.address() returns null before the 'listening' event has been emitted or after calling server.close().

Since we're listening on a port instead of a [pipe](https://learn.microsoft.com/en-us/windows/win32/ipc/named-pipes) or a [Unix domain socket](https://en.wikipedia.org/wiki/Unix_domain_socket), we can at least be sure that it'll at least be of type `AddressInfo | null`

IDEALLY, `makeRequest` should only be called when the server is listening. The server starts listening before every test (as defined in `beforeEach`).

So I think the best option here would just be to reject the promise if it's a string or null. (This would be an impossible case for the test)

Something like:
```
    const address = server.address();
    if (typeof address === 'string' || address === null) {
      reject('Unexpected address value');
      return;
    }
    const url = `http://localhost:${address.port}${path}`;
```

ericcrosson-bitgo (approved):
Adds opentelemetry support for typed-express-router

Exemplary work, @starfy84!! 🚀

ericcrosson-bitgo:
Double-checked that beta and master still point to the same commit:

```
$ git log
*   a203328 (HEAD -> beta, origin/master, origin/beta, origin/HEAD, master) Merge pull request #1043 from ericcrosson-bitgo/DX-1472-chore-transfer-code-ownership-to-devex - Eric Crosson, 4 weeks ago
```

kaustubhbitgo:
@starfy84 Can we update README.md on deprecation of `onDecodeError` methods. I was adding typed routes to express but didn't realize they are deprecated when I was following README?

starfy84:
> @starfy84 Can we update README.md on deprecation of `onDecodeError` methods. I was adding typed routes to express but didn't realize they are deprecated when I was following README?

Will do! Thanks for letting me know.

PR #1055, opened 2025-09-23
fix: updated parseRequestUnion for union requests

yanxue-22:
Ticket: DX-1437

**The error happened in this case:**
1. When it encountered a union of request types, it would:
2. Look at only the first type in the union (Type A).
3. Search inside Type A for the URL parameter definitions.
If Type A didn't have the params defined (due to inconsistency in source code), the generator would immediately stop looking without checking for param specifications in the other types.

With the changes it **iterates through all types** in the union (Type A, then Type B, etc.) until it finds the first one that contains the params definition. Once it finds the parameter, it uses it for the route and stops searching, since the URL parameters should be the same for the entire route.

I also made a few new test cases that now pass. One of them were if the first schema in a union did not have params, but the rest did, it now looks through the  rest of them instead of breaking after the first.

### DIFF
--- a/packages/openapi-generator/test/openapi/union.test.ts
+++ b/packages/openapi-generator/test/openapi/union.test.ts
@@ -361,11 +361,9 @@ export const route = h.httpRoute({
   path: '/internal/api/policy/v1/{applicationName}/touchpoints/{touchpoint}/rules/evaluation',
   method: 'POST',
   request: t.union([
-    // First schema has NO path parameters - this was causing the bug
     h.httpRequest({
       body: { emptyRequest: t.boolean }
     }),
-    // Second schema HAS path parameters - these should be preserved
     h.httpRequest({
       params: {
         applicationName: t.string,
@@ -539,7 +537,7 @@ testCase(
   },
 );
 
-const REAL_WORLD_POLICY_EVALUATION_ROUTE = `
+const ROUTE_WITH_FULLY_DEFINED_PARAMS = `
 import * as t from 'io-ts';
 import * as h from '@api-ts/io-ts-http';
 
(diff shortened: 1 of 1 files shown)


packages/openapi-generator/test/openapi/union.test.ts
```diff
@@ -353,6 +353,365 @@ testCase('route with unknown unions', ROUTE_WITH_UNKNOWN_UNIONS, {
...
 
+const ROUTE_WITH_PATH_PARAMS_IN_UNION_NOT_FIRST = `
+import * as t from 'io-ts';
+import * as h from '@api-ts/io-ts-http';
+
+export const route = h.httpRoute({
+  path: '/internal/api/policy/v1/{applicationName}/touchpoints/{touchpoint}/rules/evaluation',
+  method: 'POST',
+  request: t.union([
+    // First schema has NO path parameters - this was causing the bug
```
shenbenson, line 364: Do we want to keep these comments around? It doesn't mean anything without context

packages/openapi-generator/test/openapi/union.test.ts
```diff
@@ -353,6 +353,365 @@ testCase('route with unknown unions', ROUTE_WITH_UNKNOWN_UNIONS, {
...
+        },
+      },
+    },
+    components: {
+      schemas: {},
+    },
+  },
+);
+
+const REAL_WORLD_POLICY_EVALUATION_ROUTE = `
```
shenbenson, line 542: I think this variable name can be better

shenbenson (commented):
A few small comments, but looks good overall

shenbenson (approved):

PR #1109, opened 2026-01-15
fix: handle multibyte characters in descriptions

starfy84:
Ticket: DX-2788

This commit handles multibyte UTF-8 characters in descriptions. SWC uses UTF-8 byte offsets, while JavaScript uses character offsets. This causes a gradual offset in indices.


ericcrosson-bitgo (approved):
Thanks, @starfy84 !

PR #1111, opened 2026-01-16
fix: handle multi-byte characters

starfy84:
Ticket: DX-2800

This commit handles arbitrary sized characters. SWC uses byte offsets, while JS uses character offsets. This results in an index drift. #1109 had an initial fix, but upon adding more tests, it seemed that the fix was incomplete. This commit should pass the additional tests


ericcrosson-bitgo (approved):
Looks great, thanks @starfy84

PR #1113, opened 2026-01-16
feat: remove @preserveLineBreaks tag

starfy84:
Ticket: DX-2755

This commit removes the support for the @preserveLineBreaks tag. We will
no longer support this tag as line breaks are now preserved by default

BREAKING CHANGE: `@preserveLineBreaks` is no longer a supported tag.
Formatting preservation is now the default behavior, rendering the
`@preserveLineBreaks` tag obsolete. Please remove usages of
`@preserveLineBreaks` in your JSDoc comments


shenbenson (approved):
🧹

PR #1118, opened 2026-03-20
feat: add @public tag for x-internal: false

starfy84:
- @public on routes emits x-internal: false
- @public on schema fields/parameters emits x-internal: false
- Using both @public and @private on the same item throws an error
- Added public to known tags so it doesn't appear in x-unknown-tags


shenbenson (approved):
lgtm

PR #1128, opened 2026-04-28
chore: remove dead path substitution code in requestForRoute

oksanaphmn:
The computed path variable was never used, requestFactory already handles path parameter substitution internally via substitutePathParams.


ericcrosson-bitgo (approved):
Great! Thanks for the contribution @oksanaphmn

